PR #3, opened 2019-09-13
upload pipeline - Fix #1

chuckwagoncomputing:
Turns out the problem I was struggling with was a simple problem caused by me not reading the GH actions documentation well enough - the API key should be accessed as secrets.NUGET_API_KEY, not just NUGET_API_KEY.
The error message was terribly misleading, but I can see how the dotnet command line tool's command line parser could have made that error.


chuckwagoncomputing:
Oops, I PR'd this to the wrong repo.

PR #5, opened 2019-09-19
githubActions: improve versioning

knocte:
Just using the git hash could not be enough for proper version
sorting (last version pushed to nuget should be the first one to
be listed in the nuget website).

We also shorten the githash to not surpass the length limit of the
version string (7 is the default short git hash length, see:
https://stackoverflow.com/a/18134919/544947).


knocte:
@joemphilips before merging this, can you add the NUGET_API_KEY secret on your repo so that it uploads every time you push please?

joemphilips:
Thanks. I added NUGET_API_KEY .

PR #6, opened 2019-09-19
branch for fix failing test in azure

joemphilips:


joemphilips:
It seems that package is published from every branch. Is this intended behavior? I think it should be published from only master. @knocte

knocte:
> It seems that package is published from every branch.

That shouldn't be the case, the dotnet nuget push command is wrapped inside an `if [ $GITHUB_REF == "refs/heads/master" ]; then` statement.

joemphilips:
Closing this for now.  There is no need to do anything in azure-pipeline if we could finish everything on github actions in #7

PR #9, opened 2019-09-19
DotNetLightning.Core: lower .NETCore version requirement

knocte:
This doesn't require .NETCore 3.0 (which is actually not stable yet).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,24 @@ jobs:
     - uses: actions/checkout@v1
     - name: Install wget
       run: sudo apt install -y wget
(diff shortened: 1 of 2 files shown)


joemphilips:
Why this is necessary? I don't see any downside for supporting both.

knocte:
It's a transition commit until I actually convert it to NetStandard.

knocte:
(And anyway you're not using any .NETCore 3.0 features so the lower you can support, the better; as you will have wider target audience; since .NETCore 3.0 apps will be compatible with 2.x.)

knocte:
Superseded by https://github.com/joemphilips/DotNetLightning/pull/11

PR #10, opened 2019-09-19
Stop using NBitcoin fork

knocte:
NOTE: do not merge yet, somehow tests not passing (https://github.com/knocte/DotNetLightning/runs/228729412), help.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,6 @@ jobs:
     - name: Fetch requisite libraries
       run: |
         cd $GITHUB_WORKSPACE/src/DotNetLightning.Core
-        dotnet add package -v 4.1.2.33-joemphilips014 -s "https://www.myget.org/F/joemphilips/api/v3/index.json" NBitcoin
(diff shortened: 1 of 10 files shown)


joemphilips:
I cannot see the test log again. :(

knocte:
I sent you an invite to my fork, maybe this way you can see it?

knocte:
Wait a sec, I can see the log even if I'm logged out, just click here: https://github.com/knocte/DotNetLightning/runs/228729412

joemphilips:
When I use version  `4.1.2.33`  of NBitcoin, (which is the original version of my fork.) . This error does not happen.
So I guess it is a bug in NBitcoin side. (Or at least breaking change of TransactionBuilder's behavior.)

I will try to create a PR for NBitcoin from now on. If you are in a hurry, use the version 4.1.2.33 and ignore tests which rely on https://github.com/MetacoSA/NBitcoin/pull/742

joemphilips:
It worked when I changed the order of the opration of tx builder.
It seems that NBitcoin has become more strict about it.
(This is one reason I'm thinking to create my own version of TransactionBuilder. There are many conventions like this that I have to be aware of when using it.)

anyway. now tests are passing

PR #11, opened 2019-09-22
.NETStandard for the Core lib/nuget

knocte:
This way DotNetLightning can be consumed by platforms
different than .NETCore (e.g. Xamarin frontends, Mono...).


knocte:
This way DotNetLightning can be consumed by platforms
different than .NETCore (e.g. Xamarin frontends, Mono...).

PR #15, opened 2019-09-24
ci: update .NETCore 3.0 version from preview9 to final

knocte:
(Also use the official installation way via `apt` instead of `dpkg`.)


joemphilips:
Looks much nicer

PR #17, opened 2019-10-01
Create channel actor

joemphilips:
This is quite big change for the Infrastructure API. So I will leave a PR here for the sake of my memorandum.

Most notable change here is that Peer and Channel will both have same event-sourcing (executeCommand -> applyEvent) workflow. so that all domains can be handled by the same structure.

For example for the domain `Foo` , following classes will always be there.

1. `FooManager`
  * responsible for 1.  managing the lifetime of actors. 2. listening to events caused by other domains.
2. `FooActor`
  * responsible for running event loop for specific domain object. Every input should be done by `FooCommand`
3. `Foo`
  * The actual domain object, which is coupled with `executeCommand` and `applyEvent`

In this wayl we can assure that every changes to the domain object is thread safe and at the same time be the most performant.

Right now the only domains are `Channel` and `Peer` , but e.g. `Routing` will have the same structure.


knocte:
@ysangkok FYI^

PR #20, opened 2019-10-18
Reference Microsoft.NET.Test.Sdk

ysangkok:
Necessary to run test using "dotnet test". If I don't do this, the project is simply skipped (NET Core 3).


joemphilips:
Gosh I did not notice this since I was running manually with `dotnet run --project`

knocte:
GithubActions CI was using `dotnet test`, why wasn't it failing there then?

Maybe because of Linux/Windows differences? sigh

joemphilips:
It was just ignoring the test. What a shame of me. Partly because it has been printing all warning message.

Anyway, now it is fixed.

PR #21, opened 2019-10-25
Opt-in (non-default) BouncyCastle support

ysangkok:
Hi Joe,

As the PR title mentions, this allows for using BouncyCastle instead of Secp256k1 and NSec. There are tests for the NSec replacement parts.

If you accept this, I can add you as an owner of the DotNetLightning package on nuget, you would have to create a new API key that can upload both packages (not just one).

Let me know what you think!

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,8 @@ That is, it will read in following order.
 
 ## Supported platforms
 
-It should work on every OS if we build Secp256k1 for every platform (right now we only provide Linux binaries).
+* By default, the solution is built with dependencies on Secp256k1.Net and NSec (which have native dependencies that need to be built for every platform, and we right now only provide Linux binaries). This version is published as `DotNetLightning.Core` on Nuget.
+* For a more portable version (but possibly less performant) you can compile with BouncyCastle support, passing /p:BouncyCastle=True to your build. This version is published as `DotNetLightning` on Nuget.
 
 ## TODO
 
--- a/src/DotNetLightning.Core/Crypto/CryptoUtils.fs
+++ b/src/DotNetLightning.Core/Crypto/CryptoUtils.fs
@@ -88,10 +88,10 @@ type internal BouncySecp256k1() =
     let hex = NBitcoin.DataEncoders.HexEncoder()
     let params: Org.BouncyCastle.Asn1.X9.X9ECParameters = Org.BouncyCastle.Asn1.Sec.SecNamedCurves.GetByName "secp256k1"
     let ecParams = ECDomainParameters(params.Curve, params.G, params.N, params.H)
-    let bigint (x: byte[]) = Org.BouncyCastle.Math.BigInteger(1, x)
+    let bcBigint (x: byte[]) = Org.BouncyCastle.Math.BigInteger(1, x)
     let tweakKey (op: Op) (tweak: ReadOnlySpan<byte>) (keyToMutate: Span<byte>) =
-        let k = bigint <| keyToMutate.ToArray()
-        let tweakInt = bigint <| tweak.ToArray()
+        let k = bcBigint <| keyToMutate.ToArray()
+        let tweakInt = bcBigint <| tweak.ToArray()
         let tweaked = match op with
                       | Mul -> k.Multiply tweakInt
                       | Add -> k.Add tweakInt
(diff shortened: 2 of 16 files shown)


README.md
```diff
@@ -8,7 +8,8 @@ That is, it will read in following order.
 
 ## Supported platforms
 
-It should work on every OS if we build Secp256k1 for every platform (right now we only provide Linux binaries).
+* By default, the solution is built with dependencies on Secp256k1.Net and NSec (which have native dependencies that need to be built for every platform, and we right now only provide Linux binaries). This version is published as `DotNetLightning.Core` on Nuget.
+* For a more portable version (but possibly less performant) you can compile with BouncyCastle support, passing /p:BouncyCastle=True to your build. This version is published as `DotNetLightning` on Nuget.
```
joemphilips: I think the package name is confusing. My suggestion is

* for libsodium/nsec version ... `DotNetLightning.Core` or `DotNetLightning.Core.NSec` (or both by alias if possible )
* for BouncyCastle version ... `DotNetLightning.Core.BouncyCastle`

src/DotNetLightning.Core/Crypto/CryptoUtils.fs
```diff
@@ -1,45 +1,208 @@
...
+            use chachaKey = NSec.Cryptography.Key.Import(chacha20AD, keySpan, blobF)
+            chacha20AD.Encrypt(chachaKey, &nonce, ad, plainText)
+#else
+type internal Op = Mul | Add
+
+type internal BouncySecp256k1() =
+    let hex = NBitcoin.DataEncoders.HexEncoder()
+    let params: Org.BouncyCastle.Asn1.X9.X9ECParameters = Org.BouncyCastle.Asn1.Sec.SecNamedCurves.GetByName "secp256k1"
+    let ecParams = ECDomainParameters(params.Curve, params.G, params.N, params.H)
+    let bigint (x: byte[]) = Org.BouncyCastle.Math.BigInteger(1, x)
```
joemphilips: nit: rename to `bcBigint` so that readers wont confuse with native bigint?

src/DotNetLightning.Core/DotNetLightning.Core.fsproj
```diff
@@ -2,11 +2,31 @@
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
     <TargetsForTfmSpecificBuildOutput>$(TargetsForTfmSpecificBuildOutput);CopyProjectReferencesToPackage</TargetsForTfmSpecificBuildOutput>
-    <OtherFlags>$(OtherFlags) --warnon:1182</OtherFlags>
   </PropertyGroup>
+  <Choose>
+    <When Condition="'$(BouncyCastle)'!='' AND '$(BouncyCastle)'!='false'">
```
joemphilips: Why don't just `'$(BouncyCastle)'=='true'` ?

tests/DotNetLightning.Core.Tests/EncryptDecrypt.fs
```diff
@@ -0,0 +1,29 @@
+module EncryptDecryptTest
```
joemphilips: Would you mind adding property test for bidirectionally converting random byte vector and asserting that it won't change?
i.e. `Expect.equal (randomBytes |> encryptWithAD |> decryptWithAd) randomBytes` (Some arguments omitted )

It should be not that hard.

ysangkok:
> Would you leave here the comment explaining why this is important ?

Andrew and I will have an easier time using DotNetLightning in Geewallet with no native dependencies. Just trying to use DotNetLightning on NET Framework seems to break right now. This should enable using it on Xamarin and Framework.

knocte:
> Just trying to use DotNetLightning on NET Framework seems to break right now. This should enable using it on Xamarin and Framework.

Yes. Joe if you're curious, look at this: https://github.com/ektrah/nsec/issues/23

> I think the package name is confusing

I think the current naming makes sense if you think about it this way: "Core" suffix suggests more alignment with "Bitcoin Core" because the latter uses Secp256k1. Anyway, I thought your main goal for DotNetLightning was to use it as a daemon (not as a library exposed in nuget), so assumed any naming wrt nuget wouldn't bother you. It's likely geewallet will be the only consumer of DNL.Core lib for a long time IMO.

ysangkok:
@joemphilips Ok, I added a commit with your suggestions. If you don't want two commits, I believe Github can squash on merge now.

knocte:
Thanks for merging! The build failed because this needs to be done first:

> I can add you as an owner of the DotNetLightning package on nuget, you would have to create a new API key that can upload both packages (not just one).

@ysangkok can you add Joe as owner please? I'm having some signing problems on nuget.org myself.

PR #22, opened 2019-11-06
Core: LN.Channel: Use correct keys and to_self_delay for initial ctx

ysangkok:
I discovered this because the remote couldn't validate the commitment signature we send in funding_created. Tested with Electrum and LND on regtest.

### DIFF
--- a/src/DotNetLightning.Core/LN/Channel.fs
+++ b/src/DotNetLightning.Core/LN/Channel.fs
@@ -160,6 +160,7 @@ module Channel =
                               (initialFeeRatePerKw: FeeRatePerKw)
                               (fundingOutputIndex: TxOutIndex)
                               (fundingTxId: TxId)
+                              (pcp: PubKey)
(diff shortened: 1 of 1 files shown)


knocte:
But tests break.

ysangkok:
@joemphilips Do you think the test breakage is because of this PR? It passes locally for me, I think the test might be flaky. Especially because it mentions "timeout".

knocte:
BTW, ctx means commitment transaction? or context? I'd prefer to have less abbreviations :)

ysangkok:
@knocte it depends on ctx :P but noted, will try to use more descriptive variable names.

knocte:
Please rename :) BTW, moving forward, don't propose PRs from the master branch of any fork, otherwise the BC nuget package gets published from an unmerged source.

ysangkok:
All right, addressed your comments and opened new PR from right branch.

PR #23, opened 2019-11-06
Core: LN.Channel: Use correct keys and to_self_delay for initial commitment transactions

ysangkok:
I discovered this because the remote couldn't validate the commitment signature we send in funding_created. Tested with Electrum and LND on regtest.

<s>The tests may fail, but I think that is unrelated.</s> (EDIT: Was not unrelated, tests fixed) But if you have any suggestions I'd be happy to hear them.

### DIFF
--- a/src/DotNetLightning.Core/LN/Channel.fs
+++ b/src/DotNetLightning.Core/LN/Channel.fs
@@ -160,6 +160,8 @@ module Channel =
                               (initialFeeRatePerKw: FeeRatePerKw)
                               (fundingOutputIndex: TxOutIndex)
                               (fundingTxId: TxId)
+                              (perCommitmentPoint: PubKey)
+                              (secpContext: ISecp256k1)
                               (n: Network): RResult<CommitmentSpec * CommitTx * CommitmentSpec * CommitTx> =
             let toLocal = if (localParams.IsFunder) then fundingSatoshis.ToLNMoney() - pushMSat else pushMSat
             let toRemote = if (localParams.IsFunder) then pushMSat else fundingSatoshis.ToLNMoney() - pushMSat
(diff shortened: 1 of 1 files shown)


knocte:
Tests still red, are you sure it's a flaky test? As you say there's some timeout, try increasing it? @joemphilips any input is appreciated!

knocte:
@ysangkok nit: the names `pcp` and the abbreviature `PK` still doesn't click with me :P the latter is still ambiguous between "private key" and "public key", and the former doesn't match properly to `PubKey` type unless I'm missing something

joemphilips:
Thanks, sorry that I haven't finished debugging this kind of obvious bugs. (There should be more TBH)

I think the test failure is NOT caused by some sort of flakiness.
I can reproduce on my local machine. It is weird that you could not reproduce on your side.

I don't know the exact cause of the bug. But I think you must generate other keys as well. It seems eclair has changed its way of creating commit tx according to BOLT3 after I have ported it.

See the corresponding code in the eclair. You see it has changed the way creating txs.
https://github.com/ACINQ/eclair/blob/9d95b55e97f9b2abaa74c4c95f0003fd0742e51d/eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala#L255

So hopefully following the eclairs change will fix the problem.
(I will try fixing it by myself if you could not reproduce the failure in local.)

knocte:
@joemphilips but the interesting thing is that the tests pass when run with BouncyCastle.

joemphilips:
> @joemphilips but the interesting thing is that the tests pass when run with BouncyCastle.

hmm let me check on my local. I've never done conditional compilation so might take little time to figure out whats going on.

joemphilips:
@knocte It is because "Run Tests(BouncyCastle)" step in CI is just executing the test only for `DotNetLightning.Core` (which is fine. Because the conditional compilation is only for this package).

knocte:
oh shit good catch! @ysangkok I guess we should move thee running of thee tests to after the whole solution gets compiled, and if they pass, then upload the nuget package that was generated before.

ysangkok:
@joemphilips , I wasn't running the Infrastructure tests, that's why I didn't see this.

I have fixed the issue (was using the wrong per-commitment-point), and I have expanded the variable names.

The tests now pass.

knocte:
Great! @joemphilips ideally, merge #25 before this one.

joemphilips:
Maybe this needs rebase so that we are sure it is passing the new CI pipeline in #25 ? I'm not sure.

ysangkok:
rebased

ysangkok:
tests pass: https://github.com/nblockchain/DotNetLightning/commit/687746cdd95379577510206859280b5489ca3f10/checks

PR #24, opened 2019-11-07
Improve channel debuggability

ysangkok:
Currently, it is hard to know where in Channel.fs the channel closing
was started. Since exceptions are never raised, but only constructed,
the StackTrace member is empty. This change passes the line numbers
explicitly into the exception, so that debugging because easier, but
the API remains functional.

### DIFF
--- a/src/DotNetLightning.Core/LN/Channel.fs
+++ b/src/DotNetLightning.Core/LN/Channel.fs
@@ -42,11 +42,11 @@ type Channel = {
             }
         static member CreateCurried = curry9 (Channel.Create)
 
-type ChannelError =
-    | Ignore of string
-    | Close of string
+type CloseChannelException(msg: string, stackTrace: string) =
+    inherit System.Exception(msg)
+    override this.StackTrace = stackTrace
 
-exception ChannelException of ChannelError
+exception IgnoredChannelException of string
 
 module Channel =
     /// represents the user has than something wrong with this library
@@ -57,14 +57,11 @@ module Channel =
         msg |> APIMisuseError |> RRApiE
 
     /// Represents the error that something user can not control (e.g. peer has sent invalid msg).
-    let private RRChannelE(ex: ChannelError) =
-        RResult.rexn (ChannelException(ex))
-
-    let private RRClose(msg: string) =
-        RRChannelE(ChannelError.Close(msg))
+    let private RRClose (line: string) (msg: string) =
+        RResult.rexn <| CloseChannelException(msg, (__SOURCE_FILE__ + ": " + line))
 
     let private RRIgnore(msg: string) =
-        RRChannelE(ChannelError.Ignore(msg))
+        RResult.rexn <| IgnoredChannelException msg
 
     let private hex = NBitcoin.DataEncoders.HexEncoder()
     let private ascii = System.Text.ASCIIEncoding.ASCII
@@ -80,7 +77,7 @@ module Channel =
 
     let private checkOrClose left predicate right msg =
         if predicate left right then
-            sprintf msg left right |> RRClose
+            sprintf msg left right |> RRClose __LINE__
         else
             Good()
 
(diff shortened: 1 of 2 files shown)


src/DotNetLightning.Infrastructure/ChannelActor.fs
```diff
@@ -30,16 +30,20 @@ type ChannelActor(nodeParams: IOptions<NodeParams>,
...
-                log.LogWarning(msg)
+            | RBad.Exception(exp) ->
+                match exp with
+                | :? CloseChannelException as closeExp ->
+                    let closeCMD =
+                        let spk = _nodeParams.ShutdownScriptPubKey |> Option.defaultValue (keysRepository.GetShutdownPubKey().WitHash.ScriptPubKey)
+                        ChannelCommand.Close({ CMDClose.ScriptPubKey = Some spk })
+                    log.LogError(sprintf "Closing a channel for a node (%A) due to a following error:" (this.Channel.RemoteNodeId))
+                    log.LogError(sprintf "Exception message: %s" closeExp.Message)
+                    log.LogError(sprintf "Stack: %s" closeExp.StackTrace)
```
knocte: in .NET, printing `ex.ToString()` prints both the message and the stacktrace (and the exception type), please use this instead of the two lines above

src/DotNetLightning.Core/LN/Channel.fs
```diff
@@ -42,11 +42,11 @@ type Channel = {
             }
         static member CreateCurried = curry9 (Channel.Create)
 
-type ChannelError =
-    | Ignore of string
-    | Close of string
+type CloseChannelException(msg: string, stackTrace: string) =
+    inherit System.Exception(msg)
+    override this.StackTrace = stackTrace
```
knocte: I've never seen this weird pattern used in .NET; if you raise an exception and capture it with a `try` block you will achieve getting line numbers and stacktraces automatically

ysangkok: the exception is never raised, it is simply constructed. That's why the stack trace will be empty.

knocte: right, so why not raise it instead?

ysangkok: That is a larger refactor. If exceptions were used, there would be no need for the RBadTree and such, I guess. I don't know the motivation for avoiding exceptions, maybe @joemphilips can chime in.

knocte: Good point, let's hear Joe.

joemphilips: Well, I just wanted to make it explicit that the function is not safe from the callers point of view.
(I usually don't prefer throwing exception because consumer of the api will often forgets handling the exception.) So it is just typical motivation for using Result.

But you are right that debugging is pain when we use `RResult` naively. 

I think this fix is good enough for now. But ideally we should avoid exposing RResult to the public api all together and prepare exception type for each domain `ChannelError` `PeerError`
And each of them will hold stacktrace info (or some sort.)

joemphilips: > prepare exception type for each domain `ChannelError` `PeerError`

To make myself clear, this means model error type by DU (instead of exception) and hold stacktrace-like info in its type signature.

ysangkok: all right. I think exceptions don't mix well with functional programming because they are not reflected in the type signatures, as you mention. With a sufficiently granular DU maybe we can forget about source locations because there will only be a single hit when you grep for it. Anyway,  how do I proceed with this PR?

src/DotNetLightning.Infrastructure/ChannelActor.fs
ysangkok: My issue with `ToString` is that it doesn't have any words between those three components, only punctuation and whitespace. Since some of the errors in DotNetLightning span multiple lines, it becomes hard to see the structure. By using words it is easy to grep for e.g. `Stack:` and you will find this place.

src/DotNetLightning.Core/LN/Channel.fs
knocte: wouldn't sufficient logging in case of errors be enough? so long as you use different logging messages for each location

joemphilips: just a sec in will review more closely in an hour

joemphilips: Yes, lets keep using `ChannelError` instead of channel exception.
Using exception in this way does not seem to help.

You can change the signature as you want, adding file number field to ChannelError seems not bad for improving debugging experience.

and as @knocte suggested, logging is a good option too.

So you can do either of
1. use ChannelError and change its signature to `lineNumber: string * msg : string` for each case.
2. do nothing (close PR)
3. try refactoring out RResult and have more granular cases for ChannelError. (This is a big change and I might do it by myself later but I'd appreciate if you try it.)

src/DotNetLightning.Infrastructure/ChannelActor.fs
knocte: note that ToString() also walks the innerException tree (see some of my past nightmare stories, like this one: https://github.com/lefthandedgoat/canopy/commit/e8fea28a8771d8b38c14a32ec05cc76d3907ded8 )

knocte:
@ysangkok so, now that Joe pushed 75a19767fe7c4bf23a7bf878906a066867f7fe03 , can this be closed?

ysangkok:
@knocte Yes.

PR #25, opened 2019-11-07
CI: Only publish if tests pass on both build variants

ysangkok:
<s>Before this commit, we would run all steps unconditonally.</s> EDIT: Not true, I was mistaken. But we would publish BouncyCastle before seeing if Infrastructure builds and passes the test.

Since the native test run is running Infrastructure tests that
are not run in the first go, we wait with publishing the Bouncy-
Castle variant until the native tests have run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,52 +5,42 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
     - uses: actions/checkout@v1
-    - name: Install wget and configure Microsoft apt repository
-      run: |
-        sudo apt install -y wget apt-transport-https
-        wget -q https://packages.microsoft.com/config/ubuntu/18.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb
-        sudo dpkg -i packages-microsoft-prod.deb
-        sudo add-apt-repository universe
-        sudo apt update
-    - name: install .Net Core 2.1
-      run: |
-        sudo apt install dotnet-sdk-2.1
-    - name: Nuke NSec and Secp256k1.Net from source tree to make sure we don't depend on them
+    - name: Remove NSec and Secp256k1.Net from source tree to make sure we don't depend on them
       run: rm -rf src/NSec src/Secp256k1.Net
+    - name: Purge NET Core 3 (v2.x will be still available)
+      run: sudo apt purge dotnet-{sdk,runtime,runtime-deps}-3.0
     - name: Run tests (BouncyCastle)
       run: |
         dotnet test tests/DotNetLightning.Core.Tests -p:BouncyCastle=True
     - name: Package (BouncyCastle)
       run: |
         dotnet pack src/DotNetLightning.Core -p:Configuration=Release -p:Version=1.1.0-date`date +%Y%m%d-%H%M`.git-`echo $GITHUB_SHA | cut -c 1-7` -p:BouncyCastle=True
-    - name: Upload nuget (BouncyCastle)
-      run: |
-        cd $GITHUB_WORKSPACE/src/DotNetLightning.Core
-        if [ ${{ secrets.NUGET_API_KEY }} ] && [ $GITHUB_REF == "refs/heads/master" ]; then
-            dotnet nuget push ./bin/Release/DotNetLightning*.nupkg -k ${{ secrets.NUGET_API_KEY }} -s https://api.nuget.org/v3/index.json
-        fi
     - name: install .Net Core 3.0
       run: |
         sudo apt install dotnet-sdk-3.0
-    - name: Fetch requisite libraries
+    - name: Run Infrastructure tests on BouncyCastle
+      run: |
+        dotnet test tests/DotNetLightning.Infrastructure.Tests -p:BouncyCastle=True
+    - name: Restore NSec and Secp256k1.Net
+      run: git checkout src/NSec src/Secp256k1.Net
+    - name: Add Secp256k1.Native package
       run: |
         cd $GITHUB_WORKSPACE/src/DotNetLightning.Core
         dotnet add package -v 0.0.5-joemphilips -s "https://www.myget.org/F/joemphilips/api/v3/index.json" Secp256k1.Native
-    - name: Restore NSec and Secp256k1.Net
-      run: git checkout src/NSec src/Secp256k1.Net
     - name: Clean to prepare for NSec build
       run: |
         dotnet clean
     - name: Run tests
       run: |
         dotnet test
-    - name: Package
+    - name: Package NSec build
       run: |
         cd $GITHUB_WORKSPACE/src/DotNetLightning.Core
         dotnet pack -p:Configuration=Release -p:Version=1.1.0-date`date +%Y%m%d-%H%M`.git-`echo $GITHUB_SHA | cut -c 1-7`
-    - name: Upload nuget
+    - name: Upload nuget packages (BouncyCastle and native)
       run: |
         cd $GITHUB_WORKSPACE/src/DotNetLightning.Core
         if [ ${{ secrets.NUGET_API_KEY }} ] && [ $GITHUB_REF == "refs/heads/master" ]; then
-            dotnet nuget push ./bin/Release/DotNetLightning.Core*.nupkg -k ${{ secrets.NUGET_API_KEY }} -s https://api.nuget.org/v3/index.json
+            dotnet nuget push ./bin/Release/DotNetLightning.1*.nupkg -k ${{ secrets.NUGET_API_KEY }} -s https://api.nuget.org/v3/index.json
+            dotnet nuget push ./bin/Release/DotNetLightning.Core.1*.nupkg -k ${{ secrets.NUGET_API_KEY }} -s https://api.nuget.org/v3/index.json
         fi
--- a/tests/DotNetLightning.Core.Tests/DotNetLightning.Core.Tests.fsproj
+++ b/tests/DotNetLightning.Core.Tests/DotNetLightning.Core.Tests.fsproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <OutputType>Exe</OutputType>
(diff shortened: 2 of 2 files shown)


knocte:
This is better than before, but is still not running the Infrastructure tests with BouncyCastle.

.github/workflows/ci.yml
```diff
@@ -41,16 +35,18 @@ jobs:
...
+        dotnet test
+    - name: Upload nuget packages (BouncyCastle and native)
       run: |
         cd $GITHUB_WORKSPACE/src/DotNetLightning.Core
         if [ ${{ secrets.NUGET_API_KEY }} ] && [ $GITHUB_REF == "refs/heads/master" ]; then
-            dotnet nuget push ./bin/Release/DotNetLightning.Core*.nupkg -k ${{ secrets.NUGET_API_KEY }} -s https://api.nuget.org/v3/index.json
+            dotnet nuget push ./bin/Release/DotNetLightning.1*.nupkg -k ${{ secrets.NUGET_API_KEY }} -s https://api.nuget.org/v3/index.json
+            dotnet nuget push ./bin/Release/DotNetLightning.Core.1*.nupkg -k ${{ secrets.NUGET_API_KEY }} -s https://api.nuget.org/v3/index.json
         fi
+      if: success()
```
knocte: Are you sure this `if` is needed? if previous steps have failed, this step is already not run

.github/workflows/ci.yml
```diff
@@ -20,15 +20,10 @@ jobs:
     - name: Run tests (BouncyCastle)
       run: |
         dotnet test tests/DotNetLightning.Core.Tests -p:BouncyCastle=True
+        dotnet test tests/DotNetLightning.Infrastructure.Tests -p:BouncyCastle=True
```
knocte: not sure this is going to work as is because AFAIR the infrastructure tests need .NETCore3.0 to be compiled, that's why I was recommending to move this step after the whole solution gets compiled (and keep the `dotnet pack` call before .NETCore3 is installed, to make sure DNL.Core is still .NETCore2x compatible).

ysangkok: actually, NETCore 3 is apparently installed before the test run starts. So should I uninstall NetCore 3 after doing native build and move BouncyCastle down below?

knocte: no, .NETCore3 is installed later, so I'm not sure how this has worked

knocte: is it because the fucking GitHub guys have decide to start installing .NETCore3 by default on the ubuntu VMs???

ysangkok: Yes, as you can see in [the docs](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/software-installed-on-github-hosted-runners#ubuntu-1804-lts) it is installed in Ubuntu 18.04.

knocte: FFS... oh well, is there an easy way to uninstall it before installing .NETCore2? with apt maybe?

ysangkok: Yes, that is what I asked you: 

> So should I uninstall NetCore 3 after doing native build and move BouncyCastle down below?

knocte: ah great yes let's do that

knocte: and nit: if you edit the PR description, commit msg should follow too :)

.github/workflows/ci.yml
```diff
@@ -5,52 +5,39 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
     - uses: actions/checkout@v1
-    - name: Install wget and configure Microsoft apt repository
+    - name: Fetch Secp256k1.Native
```
knocte: I'd prefer to build and pack BC-DNL.Core before fetching any dependencies.

ysangkok: But then we'd have to reinstall NetCore 3 because we have removed it. What is the advantage of building BouncyCastle first? I am running `dotnet clean` and removing the native dependency.

knocte: Reinstalling .NETCore3 is fine. I'm just paranoid of bringing a binary blob (the one from myget) before building our nuget package.

.github/workflows/ci.yml
```diff
@@ -5,52 +5,39 @@ jobs:
...
-        sudo dpkg -i packages-microsoft-prod.deb
-        sudo add-apt-repository universe
-        sudo apt update
-    - name: install .Net Core 2.1
-      run: |
-        sudo apt install dotnet-sdk-2.1
     - name: Nuke NSec and Secp256k1.Net from source tree to make sure we don't depend on them
       run: rm -rf src/NSec src/Secp256k1.Net
+    - name: Remove NET Core 3
+      run: sudo apt remove dotnet-{sdk,runtime,runtime-deps}-3.0
```
knocte: if after removing .NETCore3, without installing .NETCore2.x, you can still run the tests, it means dotnetCore wasn't really uninstalled properly?

tests/DotNetLightning.Core.Tests/ClaimReceivedHTLCTests.fs
```diff
@@ -13,12 +13,22 @@ let utf8 = UTF8Encoding()
...
+  CommitKey: Key;
+  FinalKey: Key;
+  CommitPubKey: PubKey;
+  finalPubKey: PubKey;
+  R: byte[];
+  RHash: byte[];
+  RevokeCommit: byte[];
+  RevokeCommitHash: byte[]
+}
+let alice: Alice =
```
knocte: is this change related to CI?

tests/DotNetLightning.Core.Tests/DotNetLightning.Core.Tests.fsproj
```diff
@@ -1,9 +1,10 @@
 <Project Sdk="Microsoft.NET.Sdk" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
     <RootNamespace>DotNetLightning.Tests</RootNamespace>
     <GenerateProgramFile>false</GenerateProgramFile>
+    <DotnetFscCompilerPath></DotnetFscCompilerPath>
```
knocte: empty tag, maybe better remove it

.github/workflows/ci.yml
ysangkok: you can have multiple versions and NETCore installed, and the images have what looks like pretty much all of them installed. So why do you think it should break?

tests/DotNetLightning.Core.Tests/ClaimReceivedHTLCTests.fs
ysangkok: oops, i thought it was, but it seems that is just a local problem. i removed this.

.github/workflows/ci.yml
knocte: ah really? such a mess! (then make the name be `Remove .NETCore v3 (v2.x will be still available)`)

knocte:
Looks good! @joemphilips ideally, merge this before merging #23

.github/workflows/ci.yml
```diff
@@ -5,34 +5,20 @@ jobs:
...
-        sudo add-apt-repository universe
-        sudo apt update
-    - name: install .Net Core 2.1
-      run: |
-        sudo apt install dotnet-sdk-2.1
-    - name: Nuke NSec and Secp256k1.Net from source tree to make sure we don't depend on them
+    - name: Remove NSec and Secp256k1.Net from source tree to make sure we don't depend on them
       run: rm -rf src/NSec src/Secp256k1.Net
+    - name: Remove NET Core 3 (v2.x will be still available)
+      run: sudo apt remove dotnet-{sdk,runtime,runtime-deps}-3.0
```
knocte: oh, forgot, it's better to use `purge` than `remove`

knocte:
Wait, the last commit as it is, is not making the integration tests run with BC, or am I missing something?

.github/workflows/ci.yml
ysangkok: that only makes a difference for configuration files, and i don't even know if Net Core 3 has any. do you really want me to change this?

.github/workflows/ci.yml
```diff
@@ -44,13 +30,15 @@ jobs:
     - name: Run tests
       run: |
         dotnet test
+        dotnet test -p:BouncyCastle=True # we did not run the non-Core tests on BouncyCastle previously
```
knocte: nit: this comment might be better to put it on the commit message :)

knocte: rather put a separate step? so that if it fails, it's more clear if the failure is BC or nonBC

knocte:
:+1:

knocte:
:+1:

joemphilips:
> Wait, the last commit as it is, is not making the integration tests run with BC, or am I missing something?

I first thought running integration tests with BouncyCastle is NOT necessary. But hell yeah why not, more tests are more good.  @ysangkok Would you mind to change it?

ysangkok:
@joemphilips it is now

.github/workflows/ci.yml
```diff
@@ -5,52 +5,42 @@ jobs:
...
-        if [ ${{ secrets.NUGET_API_KEY }} ] && [ $GITHUB_REF == "refs/heads/master" ]; then
-            dotnet nuget push ./bin/Release/DotNetLightning*.nupkg -k ${{ secrets.NUGET_API_KEY }} -s https://api.nuget.org/v3/index.json
-        fi
     - name: install .Net Core 3.0
       run: |
         sudo apt install dotnet-sdk-3.0
-    - name: Fetch requisite libraries
+    - name: Restore NSec and Secp256k1.Net
+      run: git checkout src/NSec src/Secp256k1.Net
+    - name: Run all tests on BouncyCastle
```
knocte: one last nitpick! you could put this step before the "Restore NSec" step

ysangkok:
ok all comments addressed and latest version passes tests. can't run all tests without having NSec of course, so only running Infra in the second go on BC.

PR #27, opened 2019-11-14
Allow comparing NodeId with incompatible type

ysangkok:


src/DotNetLightning.Core/Utils/Primitives.fs
```diff
@@ -156,7 +156,11 @@ module Primitives =
         interface IComparable with
             override this.CompareTo(other) = if isNull other then -1 else this.Value.CompareTo((other :?> NodeId).Value)
         override this.Equals(other) =
-            if isNull other then false else this.Value.Equals((other :?> NodeId).Value)
+            if isNull other then false else
+                try
```
joemphilips: We should match against type instead of try-catching for the performance reason.
```
match other with
| :? NodeId ->
  this.Value.Equals((other :?> NodeId).Value)
| _ -> false
```

ysangkok:
Replaced by #28

PR #28, opened 2019-11-15
Core: LN.Commitments: Make sure Commitments serializes

ysangkok:
We actually tests that the WaitForFundingConfirmedData
record serializes. It contains a value of type Commitments.

### DIFF
--- a/src/DotNetLightning.Core/LN/Commitments.fs
+++ b/src/DotNetLightning.Core/LN/Commitments.fs
@@ -125,7 +125,9 @@ type Commitments = {
             let lens = Commitments.RemoteChanges_ >-> RemoteChanges.Proposed_
             Optic.map lens (fun proposalList -> proposal :: proposalList) this
 
+        [<Newtonsoft.Json.JsonIgnoreAttribute>]
         member this.IncrLocalHTLCId = { this with LocalNextHTLCId = this.LocalNextHTLCId + 1UL }
+        [<Newtonsoft.Json.JsonIgnoreAttribute>]
         member this.IncrRemoteHTLCId = { this with RemoteNextHTLCId = this.RemoteNextHTLCId + 1UL }
 
         member this.LocalHasChanges() =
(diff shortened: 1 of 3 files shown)


src/DotNetLightning.Core/LN/Commitments.fs
```diff
@@ -125,7 +125,9 @@ type Commitments = {
             let lens = Commitments.RemoteChanges_ >-> RemoteChanges.Proposed_
             Optic.map lens (fun proposalList -> proposal :: proposalList) this
 
+        [<Newtonsoft.Json.JsonIgnoreAttribute>]
         member this.IncrLocalHTLCId = { this with LocalNextHTLCId = this.LocalNextHTLCId + 1UL }
+        [<Newtonsoft.Json.JsonIgnoreAttribute>]
```
knocte: I honestly don't understand how can this compile if this project doesn't have Newtonsoft as a dependency? I'm literally scrubbing my eyes for 5 times now

knocte: oh shit, I guess it's because NBitcoin itself depends on Newtonsoft.Json I guess; anyway, you should add it as an explicit dependency as part of this PR I guess

joemphilips: I don't like the idea of using Newtonsoft explicitly in this package TBH. because dotnet is slowly moving towards `System.Text.Json` . And for `DotNetLightning.Core`, I wanted to make it json-serializer-agnostic.
That is why I was trying to have json-serialization logic in the `Infrastructure` side.

(It is true that NBitcoin already has a Newtonsoft.Json as a dependency. But we should be able to compile without it with `NOJSONNET` flag.)

Is this change really necessary? Or can't you have some kind of wrapper structure on your side?

knocte: Good points Joe. BTW maybe using this other attribute has the same effect? https://docs.microsoft.com/en-us/dotnet/api/system.nonserializedattribute?view=netframework-4.8

ysangkok:
All right, will make wrapper objects. Extracted the fixed NodeId comparison in #30

PR #30, opened 2019-11-15
Make NodeId comparable with incompatible types

ysangkok:
This version doesn't use try/with.


knocte:
@ysangkok please next time include the "why" on the commit message too, not just the "what" ;)

PR #31, opened 2019-11-16
Update dependencies, fix NSec path

ysangkok:


knocte:
What's the motivation for updating NBitcoin?

knocte:
And what's the motivation of adding the new `n` parameter? was that supposed to be in this PR?

ysangkok:
I updated because I was worried the different versions or NBitcoin were
causing problems with serialization, so I thought, why not just update all
of them in all packages. The network parameter was added because the API
changed, and the constructor with the single parameter does not exist any
more.

ysangkok:
@knocte More motivation to update NBitcoin: get https://github.com/MetacoSA/NBitcoin/pull/768 which saves us a JsonConverter.

NicolasDorier:
@knocte the new NBitcoin version has some limited support for Nullable Reference that you may like, also I removed support for some stuff that should never have worked in the first place.

PR #36, opened 2019-12-08
WIP: deprecate `RResult` and use `Result` everywhere

joemphilips:
I've been using [RResul](https://gist.github.com/mrange/1d2f3a26ca039588726fd3bd43cc8df3) everywhere for my first iteration. Which worked fine for prototyping.

However, I found out that this kind of non-strict-error-handling works only for small projects. And for the fairly big project like this, it behaves like functional version of [Pokemon anti-pattern](https://wiki.c2.com/?PokemonExceptionHandling)
and debugging has become pain. (see #24  for example).

So I'm going to refactor out RResult and use the regular Result instead.
The extension code which is necessary to handle Result efficiently is based on [FsToolKit.ErrorHanlding](https://github.com/demystifyfp/FsToolkit.ErrorHandling)


knocte:
It seems we can finally see CI statuses inside PRs! nice

knocte:
@joemphilips Good job! should @ysangkok take a look before merging this?

joemphilips:
@knocte @ysangkok give me one more day for self review.
The diff is so huge that it is hard to review, but I will appreciate if @ysangkok tries to.

joemphilips:
Finished simple self review. I will wait until around next Sunday before merging this.

PR #37, opened 2019-12-13
Reestablishment while waiting for funding tx confs

ysangkok:
Let's say we restored the channel state from disk after
a long wait for the confirmation of the funding transaction.
In this situation, we need to reestablish, and the message
to reply with is trivial since nothing has been revoked yet.

### DIFF
--- a/src/DotNetLightning.Core/Channel/ChannelTypes.fs
+++ b/src/DotNetLightning.Core/Channel/ChannelTypes.fs
@@ -297,6 +297,7 @@ type ChannelEvent =
     | Closed
     | Disconnected
     | ChannelStateRequestedSignCommitment
+    | WeReplyToChannelReestablish of msg: ChannelReestablish
 
 
 //      .d8888b. 88888888888     d8888 88888888888 8888888888 .d8888b.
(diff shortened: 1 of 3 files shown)


knocte:
@joemphilips ping

src/DotNetLightning.Core/Channel/ChannelTypes.fs
```diff
@@ -297,6 +297,7 @@ type ChannelEvent =
     | Closed
     | Disconnected
     | ChannelStateRequestedSignCommitment
+    | WeReplyToChannelReestablish of msg: ChannelReestablish
```
joemphilips: If you update the `FooEvent` in foo domain, you must also update `applyEvent` function in that domain.

ysangkok:
@joemphilips  But I don't think the channel state should be changed, so the `| otherEvent -> c` branch applies. Why does that branch exist if it is not to be used?

joemphilips:
@ysangkok you are right, I've been forgotten that branch.
I'm not sure if your logic in the `executeCommand` is correct but anyway, I will merge.
validation should probably be added after refactoring of `RResult`

PR #39, opened 2020-01-09
Use MinimumDepth from AcceptChannel message

### LINKED ISSUE #38: FundingLocked sent only when MaxMinimumDepth reached
It seems to me that DotNetLightning will only send FundingLocked once the depth from PeerLimits have been reached. But isn't that too long? If we open a channel to a counterparty that sets [minimum_depth](https://github.com/lightningnetwork/lightning-rfc/blob/master/02-peer-protocol.md#the-accept_channel-message) to a lower number, wouldn't it make sense to send FundingLocked then?

ysangkok:
When we are the funder, the counterparty determines the
MinimumDepth that the funding transaction should reach
before funding_locked.

Instead of incorrectly waiting for the limit that has been
configured in ChannelHandshakeConfig (which should be used
only for inbound channels) use the number from accept_channel,
which is now saved in RemoteParams.

Fixes #38

### DIFF
--- a/src/DotNetLightning.Core/Channel/Channel.fs
+++ b/src/DotNetLightning.Core/Channel/Channel.fs
@@ -689,7 +689,7 @@ module Channel =
                                       FirstPerCommitmentPoint = localCommitmentSecret.PubKey
                                       ShutdownScriptPubKey = cs.Config.ChannelOptions.ShutdownScriptPubKey }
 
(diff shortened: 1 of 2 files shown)


knocte:
@ysangkok Good work!

@joemphilips please review :)

joemphilips:
Looks perfect.

PR #40, opened 2020-01-13
Make CreateOutbound take an explicit private key

ysangkok:
Before this commit, a random key would be generated, which is not useful
outside testing, since users will usually reuse the same transport
encryption key for both outbound and inbound connections.

By making CreateOutbound take an explicit key, consumers of this
function can explicitly generate a key if they so desire.

Consumers in Infrastructure have been modified to now generate a key
themselves, retaining previous behaviour.

### DIFF
--- a/tests/DotNetLightning.Core.Tests/PeerChannelEncryptorTests.fs
+++ b/tests/DotNetLightning.Core.Tests/PeerChannelEncryptorTests.fs
@@ -14,8 +14,7 @@ let getOutBoundPeerForInitiatorTestVectors () =
 
     let outboundPeer =
         let ie = (Key(hex.DecodeData("1212121212121212121212121212121212121212121212121212121212121212")))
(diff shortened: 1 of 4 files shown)


tests/DotNetLightning.Core.Tests/PeerChannelEncryptorTests.fs
```diff
@@ -14,7 +14,7 @@ let getOutBoundPeerForInitiatorTestVectors () =
 
     let outboundPeer =
         let ie = (Key(hex.DecodeData("1212121212121212121212121212121212121212121212121212121212121212")))
-        PeerChannelEncryptor.newOutBound(NodeId theirNodeId)
+        PeerChannelEncryptor.newOutBound(NodeId theirNodeId, Key ())
         |> Optic.set PeerChannelEncryptor.OutBoundIE_ ie
```
joemphilips: I think this line has become useless now
also. would you pass the variable ie here instead of creating new key?

PR #42, opened 2020-01-27
Fix #41

### LINKED ISSUE #41: Unable to find package ResultUtils. No packages exist with this id in source(s): nuget.org
I suspect I am getting this error in geewallet because I am not using your feed. Would it be possible to publish this package on Nuget? Thanks.

joemphilips:
~This should solve #41~ I'm on the way of searching the best way...

### DIFF
--- a/src/DotNetLightning.Core/DotNetLightning.Core.fsproj
+++ b/src/DotNetLightning.Core/DotNetLightning.Core.fsproj
@@ -19,8 +19,8 @@
   </Choose>
   <ItemGroup>
     <ProjectReference Condition="'$(BouncyCastle)'!='true'" Include="..\Secp256k1.Net\Secp256k1.Net.csproj" PrivateAssets="all" />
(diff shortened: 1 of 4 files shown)


knocte:
I think this is the wrong approach because it would make a piece of OSS software depend on binary blobs ;-)

joemphilips:
umm yes.. I'm seeking for the right approach now...
I think [this issue](https://github.com/dotnet/cli/issues/3959) is where I have to take a look carefully

knocte:
Oh you mean like I did in c98307465f647257df1438beadb4cabc7db757f2 ?

joemphilips:
Yes, I still don't understand the detail of what the code you introduced in that commit do, but I thought adding `PrivateAssets = "all"` to every references I have will solve the problem.

But I think this is not the best way, because the consumer of this package probably want to use `ResultUtils` too.
(TBH, I don't understand what you have mentioned as a binary blobs.)

knocte:
With binary blobs I meant binaries whose source code we don't have (because at first I thought that ResultUtils came from a private feed which came from a private repo of yours, when I first read Janus' comment).

PR #43, opened 2020-01-29
Flexible ChannelReestablish

ysangkok:
Before, we could only generate a channel_establish message if we
received one. Obviously, not every node can work like this.
Furthermore, since we weren't really validating the incoming
message anyway, there was no added benefit to requiring it.

This commit makes it possible to generate a channel_reestablish
message in all states, not just when we are waiting for the funding
transaction to confirm. For example, we need to generate the
message in the Normal state very often.

### DIFF
--- a/src/DotNetLightning.Core/Channel/Channel.fs
+++ b/src/DotNetLightning.Core/Channel/Channel.fs
@@ -107,6 +107,19 @@ module Channel =
                 failwith "TODO"
             }
 
+    let makeChannelReestablish (data: Data.IHasCommitments): Result<ChannelEvent list, ChannelError> =
+        let commitmentSeed = data.Commitments.LocalParams.ChannelPubKeys.CommitmentSeed
+        let ourChannelReestablish =
+            {
+                ChannelId = data.ChannelId
+                NextLocalCommitmentNumber = 1UL
+                NextRemoteCommitmentNumber = 0UL
+                DataLossProtect = OptionalField.Some({
+                                      YourLastPerCommitmentSecret = PaymentPreimage([|for _ in 0..31 -> 0uy|])
+                                      MyCurrentPerCommitmentPoint = ChannelUtils.buildCommitmentPoint(commitmentSeed, 0UL)
+                                  })
+            }
+        [ WeSentChannelReestablish ourChannelReestablish ] |> Ok
 
     let executeCommand (cs: Channel) (command: ChannelCommand): Result<ChannelEvent list, ChannelError> =
         match cs.State, command with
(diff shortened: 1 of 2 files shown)


src/DotNetLightning.Core/Channel/Channel.fs
```diff
@@ -107,6 +107,19 @@ module Channel =
...
 
+    let makeChannelReestablish (data: Data.IHasCommitments): Result<ChannelEvent list, ChannelError> =
+        let commitmentSeed = data.Commitments.LocalParams.ChannelPubKeys.CommitmentSeed
+        let ourChannelReestablish =
+            {
+                ChannelId = data.ChannelId
+                NextLocalCommitmentNumber = 1UL
+                NextRemoteCommitmentNumber = 0UL
+                DataLossProtect = OptionalField.Some({
+                                      YourLastPerCommitmentSecret = PaymentPreimage([|for _ in 0..31 -> 0uy|])
```
knocte: what is 31?

ysangkok: 32 minus one. PaymentPreimage will check the length. The same idiom is used elsewhere, see https://github.com/joemphilips/DotNetLightning/blob/75a19767fe7c4bf23a7bf878906a066867f7fe03/tests/DotNetLightning.Core.Tests/Serialization.fs#L604

knocte: and what is 32? I mean is this a magic number?

joemphilips: It is defined in bolt-2 that `payment_preimage` must be 32 bytes.
In general, when we have some kind of secret-and-commitment scheme in bitcoin, usually 32 bytes secret is used (even though technically it can be longer)
This is because 32 bytes secret is the shortest length to assure the enough entropy for its hash.

joemphilips: But I don't understand why it is ok here to use non-random secret. Shouldn't be some kind of random bytes?

knocte: shouldn't the consumer of DotNetLightning be the one in charge to provide the randomness source instead of DNL itself?

joemphilips: Yes, that is the purpose for `IKesRepository` .

ysangkok: @knocte If no commitment has been revoked, no secret is available. The channel can be reestablished before a commitment has been revoked. A secret is revoked as part of the `revoke_and_ack` message. If we have not received that message, no secret is available. Furthermore, that code is existing, and this commit is concerned with moving it out into its own function so that we can restore channels in the Normal state also, which is necessary since Geewallet now saves channels so that they have Normal state on disk. I will replace this code with something better once the time is right. But there are bound to be other bugs concerning adding and removing commitments, so I think it is premature to worry about if this piece of code is general enough.

knocte: I was just nitpicking about magic numbers, which I fixed in https://github.com/joemphilips/DotNetLightning/pull/45 (WIP)

joemphilips: Will at least this PR is better than before. So I suppose I should just merge this and think about the plausibility of using  `[|for _ in 0..31 -> 0uy|]` as a preimage later.

joemphilips:
oops. I forgot to check the CI before merging... 
Current CI process is not good for checking a PR from forked repository... I should look for the solution.

knocte:
Yeah build is broken now.

joemphilips:
I have reverted this change on master and introduce branch protection. @ysangkok would you recreate this PR? Thanks.

PR #45, opened 2020-01-29
Payment preimage slight refactorings

knocte:
See each individual commit message for details.

### DIFF
--- a/src/DotNetLightning.Core/Utils/Primitives.fs
+++ b/src/DotNetLightning.Core/Utils/Primitives.fs
@@ -159,7 +159,10 @@ module Primitives =
                 member this.ToPubKey() =
                     this.ToPrivKey().PubKey
 
-
+    let (|PaymentPreimage|) x =
+        match x with
+        | PaymentPreimage x -> Some (x)
+        
     type ChannelId = | ChannelId of uint256 with
         member x.Value = let (ChannelId v) = x in v
 
--- a/src/DotNetLightning.Core/Utils/Primitives.fs
+++ b/src/DotNetLightning.Core/Utils/Primitives.fs
@@ -1,8 +1,11 @@
 namespace DotNetLightning.Utils
+
 open NBitcoin
 open NBitcoin.Crypto
+
 open System
 open System.Net
+open System.Linq
 
 [<AutoOpen>]
 module Primitives =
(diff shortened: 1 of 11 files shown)


src/DotNetLightning.Core/Utils/Primitives.fs
```diff
@@ -117,30 +120,43 @@ module Primitives =
...
+        member x.ToBytes(?lEndian) =
+            let e = defaultArg lEndian true
+            x.Value.ToBytes(e)
+
+        member x.GetRIPEMD160() =
+            let b = x.Value.ToBytes()
+            Crypto.Hashes.RIPEMD160(b, b.Length)
+
+
+    type PaymentPreimage(value: seq<byte>) =
```
joemphilips: I prefer to use single case union. so that it aligns well with other types defined in this file.

src/DotNetLightning.Core/Utils/Primitives.fs
```diff
@@ -117,30 +120,43 @@ module Primitives =
...
+        do
+            if value.Count() <> PaymentPreimage.LENGTH then
+                raise <| ArgumentException(sprintf "Payment preimage length should be %i" PaymentPreimage.LENGTH)
```
joemphilips: Validating the length here is not bad. in fact it is better than before that I did not have any validation.
But returning result is better. such as
```fsharp
type PaymentPreimage = private PaymentPreimage of seq[]

module PaymentPreimage =
  let (|PaymentPreimage|_) x =
    match x with
    | PaymentPreimage x -> Some x
    | _ -> None

  let TryCreate (data: seq<bytes>) =
    // validate and return result
```

knocte: I'm not familiar with this `let (|foo|) bar =` thing, what is that?

src/DotNetLightning.Core/Utils/Primitives.fs
knocte: Out of curiosity, any special reason why a single case union would be better? Just for consistency or more reasons?

joemphilips: I love automatic equality and comparison implementation in single case union.
serialization property test will be easy, for example.

Not sure that is the case for this one though.

src/DotNetLightning.Core/Utils/Primitives.fs
joemphilips: Its active pattern (may be it was `let (|foo|_|) bar =` ) it is necessary for pattern matching in the case of private constructor.

src/DotNetLightning.Core/Utils/Primitives.fs
knocte: Ah good point!

src/DotNetLightning.Core/Utils/Primitives.fs
knocte: can you point me to any docs about this?

joemphilips: https://habr.com/en/post/458730/

joemphilips: Although the example explained in this post is for non-single DU. it has an use case in single case DU when you deconstruct PaymentPreimage in function argument. e.g.

```fsharp
let (PaymentPreimage x) =
    // do something with x
```

Speaking precisely you don't need module in this case. just place active pattern under the type definition. And TryCreate can be static method instead.

knocte: it seems the types created with `type Foo = private Foo of Bar<Baz>` which later can have a `with` clause following augmentations, don't allow `let` declarations inside. Or maybe I'm missing something? So we may need a module after all.

joemphilips: What I meant is,  in this case `type PaymentPreimage` is already placed in another module, so we can define active pattern directly *next* to the PaymentPreimage itself. (otherwise if PaymentPreiamge is placed under namespace, we have to prepare another module since it is impossible to place active pattern directly under namespace.)

knocte: I'm sorry but I still don't get what you mean. Anyway, the current master branch doesn't have an active pattern on this type, so it's not being used at the moment. Either we should only add it when it's needed, or it can be added outside of this PR.

knocte: Argh now the tests don't compile, is it because I need the active pattern?

src/DotNetLightning.Core/Utils/Primitives.fs
```diff
@@ -117,31 +120,49 @@ module Primitives =
...
+
+                member this.ToPrivKey() =
+                    this.ToByteArray() |> Key
+
+                member this.ToPubKey() =
+                    this.ToPrivKey().PubKey
+
+    let (|PaymentPreimage|) x =
+        match x with
+        | PaymentPreimage x -> Some (x)
```
joemphilips: This is what I meant.

joemphilips:
I have fixed tests. will merge this if it passes CI.

joemphilips:
Code is fine and it passes the tests. But I want to tweak the settings of CI before merging. It seems not working well for PR from other repo.

PR #54, opened 2020-02-27
Fix Secp256k1 on 64bit Linux

canndrew:
The bindings to the native library functions were incorrectly using C# `uint` in place of C's `size_t`. This was causing the library to segfault on 64bit linux due to C#'s `uint` being 32bit whereas C's `size_t` is 64bit.

This PR changes these bindings to use `UIntPtr` as a closer approximation of `size_t`.


joemphilips:
Oh my I don't know why I got this wrong.
thanks @canndrew

PR #55, opened 2020-02-27
change supported feature according to our own plan

joemphilips:
My plan is to support following feature in [BOLT 9](https://github.com/lightningnetwork/lightning-rfc/blob/master/09-features.md) in a first iteration.

1. option_data_loss_protect
2. initial_routing_sync
3. option_upfront_shutdown_script
4. gossip_queries
5. var_onion_optin
6. option_static_remotekey
7. payment_secret

(It is only a plan and I might change my mind during coding.)
So I have updated the FeatureBit and its tests. next, I will prepare P2P messages necessary for those feature.


joemphilips:
Finished P2P message necessary for `gossip_queries` , I think there is no work left to do on P2P message itself (although we have to update onion hop message format so that we can support `var_onion_optin` , but this is responsibility of `Sphinx.fs`.) 
merging for now.

PR #66, opened 2020-03-04
WIP: implement routing module

joemphilips:
After spending some time searching what is the best and greatest algorithm for LN routing, I realized this is on-going research topic and no one knows what is the best.

So I switched my mind to just get shit done rather than searching for the perfect solution.

Algorithm here is mostly simply copied from eclair with one difference, eclair uses adjacent list to represent the graph structure, but here I used inductive graph (see this [paper](https://web.engr.oregonstate.edu/~erwig/papers/InductiveGraphs_JFP01.pdf) and its implementation [FSharp.FGL](https://github.com/CSBiology/FSharp.FGL) for the detail.)

Usage of FGL is seems to be overkill right now. but I believe it has benefit when maintaining or refactoring the module later.


nufsi.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/sh
+
```
knocte: just curious, what's this script for?

joemphilips: I included here by mistake. It is from [this blog](https://two-wrongs.com/referencing-nuget-packages-in-fsharpi)
I will remove from this PR later.

joemphilips:
I've changed my mind for using Inductive graph. FGL does not support the case when there are multiple edges between same pair of nodes. (Which is very likely to happen in the LN channel network.)

joemphilips:
Finished porting every tests for route calculation. I will probably merge this after few days.

knocte:
Looks like great work! @ysangkok you might want to review and/or have a look.

PR #72, opened 2020-03-09
Use the correct basepoint in makeRemoteTxs

canndrew:
The `Commitments.Helpers.makeRemoteTxs` function incorrectly uses the remote peer's `PaymentBasePoint` instead of `DelayedPaymentBasePoint` when calling `makeCommitTx`. This causes signature checks to fail.

This PR fixes this.


joemphilips:
You are correct. Thanks for finding this 🙏

PR #73, opened 2020-03-10
Fix dummy funding tx generation

canndrew:
The test suite's `ProvideFundingTx` function would randomly (50% of the time) return an incorrect funding output index. This is because NBitcoin was shuffling the funding tx's txouts, but the function would always return an output index of 0.

This fixes the problem by making `ProvideFundingTx` disable txout shuffling on the transaction builder and making it always return an output index of 1.


knocte:
Yeah what's the point of randomizing? I don't understand.

canndrew:
> Yeah what's the point of randomizing? I don't understand.

¯\_(ツ)_/¯. It does it by default though unless you explicitly turn it off. Ordering the txins/txouts would make more sense.

canndrew:
Ordering the txins/txouts lexicographically is what I meant.

PR #77, opened 2020-03-23
Change BlockHeightOffset from 16bit to 32bit

canndrew:
It's true that, as mentioned in the code, `OP_CSV` only allows 16bit offsets. However it is possible to have blocks that are more than `UINT16_MAX` blocks apart, and lightning's `accept_channel` message contains a 32bit block height offset for the `minimum_depth` field. Because of this, the `AcceptChannel` type's `MinimumDepth` field was using the 32bit `BlockHeight` type to represent a relative block height, even though `BlockHeight` is intended to represent absolute block height.

For the sake of using `BlockHeight` consistently as an absolute block height I've changed `BlockHeightOffset` to be 32bits and changed the type of `AcceptChannel.MinimumDepth` to be a `BlockHeightOffset`.


joemphilips:
@canndrew  good catch.

But I'm not sure this is a right way to go. It seems to me that loosing the restriction for `BlockHeightOffset` is not good for a security reason. (since we might create invalid CSV script or such.)

What about just using `uint32` for `minimum_depth` in `accept_channel` ? I think this is the only place in the protocol which uses `uint32` to represent relative block height. so treating the field as an exception to our type modeling makes sense to me.

canndrew:
We could also have separate `BlockHeightOffset16` and `BlockHeightOffset32` types. I prefer to use strong types whereever possible, and it's also possible in principle to have block height offsets greater than 65535 given that the block chain is already ~10x that long.

joemphilips:
> We could also have separate BlockHeightOffset16 and BlockHeightOffset32 types

Yeah that also seems ok to me.

joemphilips:
Should we close this now?

canndrew:
> Should we close this now?

Yep :)

Issue #74, opened 2020-03-11
Consider binding to rust-lightning.


joemphilips:
Recently, rust-lightning got sponsored by square crypto and it is likely to be maintained in a long term as Lightning Developer Kit.
If we are able to create a binding against rust-lightning, DotNetLightning.Core can be a thin wrapper around it and we don't have to care about nitty-gritty part like channel state management or so.

joemphilips:
@knocte taught me about @canndrew 's previous [attempt to create the bindings against rust-lightning](https://gitlab.com/nblockchain/geelightning),.
See [his comments here](https://gitlab.com/nblockchain/geewallet/-/merge_requests/72#note_279998877)

 @canndrew seems to understand about FFI much better than me, and the fact that even he decided not to go on that path made me hesitant to trying it by myself.


I may come back to this issue later if someone creates nice and stable FFI interfaces for rust-lighting, or [wasm loading](https://hacks.mozilla.org/2019/12/using-webassembly-from-dotnet-with-wasmtime/) becomes a thing.

knocte:
geelightning is not really rust-lightning related, just rust-related (but the struggles to bind it from F# still apply). But yeah, if you decide to switch DNL to be just a binding to rust-lightning, we might reconsider our options; we didn't set anything in stone yet though. Another good reading from a very senior developer trying to integrate .NET from Rust: https://ericsink.com/entries/dotnet_rust.html

joemphilips:
> geelightning is not really rust-lightning related, just rust-related 

whoops I thought it is a fork of rust-lightning since the directory structure was the same.

joemphilips:
I'm resuming this, by following the practice in [rust-csharp-ffi](https://github.com/KodrAus/rust-csharp-ffi), I think I can manage to create the FFI. It may take me some time but I want to dig in much deeper before giving it up.

joemphilips:
For someone interested, the PR is [over here](https://github.com/rust-bitcoin/rust-lightning/pull/569)

And the code for C# wrapper is [over here](https://github.com/joemphilips/NRustLightning).

knocte:
> And the code for C# wrapper is over here.

So we can close this issue?

joemphilips:
Yeah why not. I use DNL internally in my binding project. But I suppose we should keep it separated.

PR #80, opened 2020-04-08
Split core packages

joemphilips:
Right now I am testing C# bindings for rust-lightning (see #74 ),
and I want to pass/receive binary serialized data for P2P messages across ffi boundary.

Rather than re-implementing, I'd better use what I have already done in this repository.
But `DotNetLightning.Core` has too much features that I don't use for binding (e.g. Channel State handling, cryptographic ops.)
So In this PR, I will split the project into two 1. `DotNetLightning.Data` for low level primitives and 2. `DotNetLightning.Core` for other things.

It should not break the build since namespaces and visibility is not affected (I suppose).


knocte:
@joemphilips do you mind creating a "stable" branch before merging this?

joemphilips:
@knocte what do you mean? "stable" for release channel and "master" for releasing nightly version?

Then I think leaving "master"  for stable release and having "staging" branch sounds better to me.

knocte:
Sorry but I have a very strong opinion on this. 'master' should always reflect the latest work being done on a project, and other branches should be checkpoints that QA can bless independently.

knocte:
(It's how the majority of projects work. This way there's no confusion of where should devs create PRs from, for example. I can give you countless examples of projects working this way.)

joemphilips:
Ok. Do you think we should publish nuget package from both?

joemphilips:
If you have no opinion, I will probably add "-stable-" and "-master-" in the version of  the nuget package and publish both.

joemphilips:
Or maybe stop adding suffix to the version for stable release.

knocte:
> If you have no opinion

Just give me time to reply :P

> Or maybe stop adding suffix to the version for stable release.

Yeah I think this is the way, master would always push versions to nuget that have date and git hash (which are treated by default by nuget as "prereleases", just because of the fact that the contain non-numbers on them), but the stable branch ideally would have just numeric versioning (which would make nuget consider them as releases, not pre-releases).

joemphilips:
Finished stabilizing release process in #84 .
I did not introduce new `stable` branch because I thought it is not necessary. Instead, I just push new tag from local and an github action will create new release and nuget package.

I will merge this in few days if @knocte is ok with it.

knocte:
> I will merge this in few days if @knocte is ok with it.

If you merge #85 before merging this one, I'm happy.

joemphilips:
I changed my mind about this.
I will just leave it as is and use the package as a whole, rather than complicating things.

PR #85, opened 2020-04-14
Add RemoteNextCommitInfo type

canndrew:
Replace the use of `Choice1Of2` for storing the next remote commit info with a new discriminated union type which has two variants. I found that this makes the code a lot clearer and easier to understand.

### DIFF
--- a/src/DotNetLightning.Core/Channel/Commitments.fs
+++ b/src/DotNetLightning.Core/Channel/Commitments.fs
@@ -88,6 +88,30 @@ type WaitingForRevocation = {
             (fun w -> w.ReSignASAP),
             (fun v w -> { w with ReSignASAP = v })
 
+type RemoteNextCommitInfo =
+    | Waiting of WaitingForRevocation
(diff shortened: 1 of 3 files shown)


src/DotNetLightning.Core/Channel/Commitments.fs
```diff
@@ -88,6 +88,30 @@ type WaitingForRevocation = {
             (fun w -> w.ReSignASAP),
             (fun v w -> { w with ReSignASAP = v })
 
+type RemoteNextCommitInfo =
+    | WaitingForRevocation of WaitingForRevocation
```
joemphilips, line 92: It is good to have a distinct identifier for DU case and the type of data inside that case.
Since you don't have to match it by full name in that way.
In this case I think
```
| Waiting of WaitingForRevocation
```

is better.

canndrew: That's a good point. I've changed it.

PR #86, opened 2020-04-21
Make GetFailureMsgData format non-printable error messages

canndrew:
Some lightning implementations will send non-printable data in error messages. For instance outdated versions of lnd will do this:

https://github.com/lightningnetwork/lnd/commit/ff37b711c6dea62f6acb9d0aaca36417edcda820

Where these are the error codes (sent as single-byte error messages):

https://github.com/lightningnetwork/lnd/blob/9b1ecbd3fa3fd93a6d95a5f71921c73777340760/lnwire/error.go#L12

These errors need to be printed as bytes when displayed to the user, rather than interpreted as strings. Error.GetFailureMsgData will now checks whether an error message is printable before returning the message as a string and, if it's not, returns a string containing the hex-encoded binary data in the error.

### DIFF
--- a/src/DotNetLightning.Core/Serialize/Msgs/Msgs.fs
+++ b/src/DotNetLightning.Core/Serialize/Msgs/Msgs.fs
@@ -1281,7 +1281,20 @@ type ErrorMessage =
                 ls.WriteWithLen(this.Data)
 
         member this.GetFailureMsgData() =
-            System.Text.ASCIIEncoding.ASCII.GetString(this.Data)
+            let minPrintableAsciiChar = 32uy
+            let isPrintableAsciiChar (asciiChar: byte) =
+                asciiChar >= minPrintableAsciiChar
+            let isPrintableAsciiString =
+                not <| Seq.exists
+                    (fun asciiChar -> not (isPrintableAsciiChar asciiChar))
+                    this.Data
+            if isPrintableAsciiString then
+                System.Text.ASCIIEncoding.ASCII.GetString this.Data
+            else
+                Seq.fold
+                    (fun msg (asciiChar: byte) -> sprintf "%s %02x" msg asciiChar)
+                    "<error contains non-printable binary data>:"
+                    this.Data
 
 and WhichChannel =
     | SpecificChannel of ChannelId


src/DotNetLightning.Core/Serialize/Msgs/Msgs.fs
```diff
@@ -1281,7 +1281,15 @@ type ErrorMessage =
                 ls.WriteWithLen(this.Data)
 
         member this.GetFailureMsgData() =
-            System.Text.ASCIIEncoding.ASCII.GetString(this.Data)
+            let isPrintableAscii =
+                not <| Array.exists (fun asciiByte -> asciiByte < 32uy) this.Data
+            if isPrintableAscii then
+                System.Text.ASCIIEncoding.ASCII.GetString(this.Data)
```
knocte, line 1287: nit: no need for parenthesis here

src/DotNetLightning.Core/Serialize/Msgs/Msgs.fs
```diff
@@ -1281,7 +1281,15 @@ type ErrorMessage =
...
+            else
+                Array.fold
```
knocte, line 1289: nit: can you use Seq.fold instead? (equivalent of IEnumerable in the C# world) this way it will be future proof in case List is used instead of Array later on

knocte: (same above about Array.exists -> Seq.exists)

src/DotNetLightning.Core/Serialize/Msgs/Msgs.fs
```diff
@@ -1281,7 +1281,15 @@ type ErrorMessage =
                 ls.WriteWithLen(this.Data)
 
         member this.GetFailureMsgData() =
-            System.Text.ASCIIEncoding.ASCII.GetString(this.Data)
+            let isPrintableAscii =
+                not <| Array.exists (fun asciiByte -> asciiByte < 32uy) this.Data
```
knocte, line 1285: magic number: 32uy; if you put it under a constant, and choose the right name, code will be more understandable

canndrew:
I've updated the commit.

src/DotNetLightning.Core/Serialize/Msgs/Msgs.fs
```diff
@@ -1281,7 +1281,20 @@ type ErrorMessage =
...
+                asciiChar >= minPrintableAsciiChar
+            let isPrintableAsciiString =
+                not <| Seq.exists
+                    (fun asciiChar -> not (isPrintableAsciiChar asciiChar))
+                    this.Data
+            if isPrintableAsciiString then
+                System.Text.ASCIIEncoding.ASCII.GetString this.Data
+            else
+                Seq.fold
+                    (fun msg (asciiChar: byte) -> sprintf "%s %02x" msg asciiChar)
```
joemphilips, line 1295: FYI, using `StringBuilder` and `Seq.iter` is better for performance.
(I don't think the performance will ever be an issue here so it is ok.)

PR #88, opened 2020-04-29
Fix feature set calculation when deserializing init msg

canndrew:
The `global_features` and `local_features` fields of the init msg need to be ORed together to create the complete feature set. Previously they were being concatenated, resulting in an invalid feature set.

### DIFF
--- a/src/DotNetLightning.Core/Serialize/Msgs/Msgs.fs
+++ b/src/DotNetLightning.Core/Serialize/Msgs/Msgs.fs
@@ -390,10 +390,25 @@ type Init =
                 this.Features <- oredFeatures |> FeatureBit.CreateUnsafe
                 this.TLVStream <- ls.ReadTLVStream() |> Array.map(InitTLV.FromGenericTLV)
             member this.Serialize(ls) =
-                // For backwards compatibility reason, we must consider legacy `global features` section. (see bolt 1)
-                let g: byte[] = [||]
-                ls.WriteWithLen(g)
-                ls.WriteWithLen(this.Features.ToByteArray())
+                // Trim any leading zero bytes from the from of the local features
+                let localFeatures =
+                    let features = this.Features.ToByteArray()
+                    let startIndex =
+                        match Array.tryFindIndex (fun (b: byte) -> b <> 0uy) features with
+                        | Some index -> index
+                        | None -> features.Length
+                    features.[startIndex..]
+                // For legacy compatibility, the last (lowest) 13 bits get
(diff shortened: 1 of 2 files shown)


joemphilips:
I have to look into the spec.
I'm surprised that this passes tests, especially a property test.

joemphilips:
Ohh, I see. Yeah it seems fine.
I'm not sure that the legacy remote peer will accept empty global feature but that seems another story.

In that case we must resume `GlobalFeatures` field in `Init` I suppose.

joemphilips:
Would you mind adding `GlobalFeatures` field to init message in this PR? I'm starting to think this is necessary change.

joemphilips:
Actually I realized that there are no needs to hold `GlobalFeatures` field in `Init` .
We just to have to write up to 13 bits from `Features` field.

See [init deserialization part](https://github.com/rust-bitcoin/rust-lightning/blob/12e2a81e1daf635578e1cfdd7de55324ed04bd48/lightning/src/ln/msgs.rs#L855) in rust-lightning.
And [write_up_to_13](https://github.com/rust-bitcoin/rust-lightning/blob/12e2a81e1daf635578e1cfdd7de55324ed04bd48/lightning/src/ln/features.rs#L90) function.

canndrew:
I've amended `Init.Serialize` so that it sends the lower 13 bits of the feature set in `global_features` (as well as in `local_features`). I've also fixed a bug in `BitArray.Reverse` which was causing it to drop bits.

canndrew:
Unfortunately there's still a problem with this. The change to the lightning spec which removed the distinction between global and local features was only done recently, so older nodes don't yet accept the feature sets that we're sending. Specifically, @knocte has been trying to test against this node: 0382019f91e0d8a5e497849052b5bedad4329bc88ca24b5dbf499c0bea1af5b6c5@172.81.182.181:9735, but this node will only accept our init if we include `option_data_loss_protect` *only* in the `local_features` section. If it's also specified in `global_features` then the node immediately disconnects from us.

I'm not sure if this is worth fixing since we're the ones being spec-compliant here and other implementations will eventually update themselves to be compatible with us. However we could still be spec-compliant while supporting old nodes it would just require extra logic to only signal certain features in certain fields.

knocte:
> However we could still be spec-compliant while supporting old nodes it would just require extra logic to only signal certain features in certain fields.

Ok let's do this in a new PR. In the meantime, @joemphilips can this one be merged?

joemphilips:
OK.

PR #89, opened 2020-05-04
trigger CI action by push not pull_request

joemphilips:


.github/workflows/ci.yml
```diff
@@ -1,5 +1,5 @@
 name: Build DotNetLightning and deploy to NuGet
-on: [pull_request]
+on: [push]
```
knocte, line 2: @joemphilips but this now means that you will not see CI status when someone opens a PR (and then you will need to go into their branch to see if it passed or not); ideal is how it was before: both push and pull_request

knocte: see what I mean? https://github.com/joemphilips/DotNetLightning/pull/90

joemphilips: aaagh. OK.

PR #92, opened 2020-05-06
Init message legacy compatibility

canndrew:
The PR makes init message serialization give special treatment to the `var_onion_optin` feature flag. This flag (if it is present) will only be placed in the `global_features` section of the init message while other flags are only placed in the `local_features` section. This makes use compatible with legacy clients that distinguish between global and local features since `var_onion_optin` was the only global feature in the spec before the spec was changed and the two feature sets were merged.

Some other miscellaneous changes are:

I added `SetFeatureFlag` and `GetFeatureFlag` methods to the `FeatureBit` type since this made it more convenient to implement `Init.Serialize`.

I've removed the lazily-initialized `mutable bytes` field from `FeatureBit` since this made it possible for `FeatureBit.ByteArray` to become out of sync with `FeatureBit.BitArray`. `ByteArray` is now computed from the internal `bitArray` on demand.

I modified the `BitArray.ToByteArray` extension method so that it is the inverse of the `BitArray.FromBytes` extension method, rather than the two methods ordering the bits within the bytes differently. `BitArray.ToByteArray` will also return the minimal-length array required to represent the `BitArray` rather than (possibly) containing `0x00` bytes at the front (which is a protocol violation if sent over the network as a feature set).

`BitArray.Reverse` now returns a `BitArray` and is implemented using `Array.Reverse` rather than with a `for` loop.

### DIFF
--- a/src/DotNetLightning.Core/Serialize/Features.fs
+++ b/src/DotNetLightning.Core/Serialize/Features.fs
@@ -29,6 +29,11 @@ type Feature = private {
     with
     member this.MandatoryBitPosition = this.Mandatory
     member this.OptionalBitPosition = this.Mandatory + 1
+    member this.BitPosition(support: FeaturesSupport) =
+        match support with
+        | Mandatory -> this.MandatoryBitPosition
+        | Optional -> this.OptionalBitPosition
+
     override this.ToString() = this.RfcName
 
     static member OptionDataLossProtect = {
@@ -89,7 +94,7 @@ module internal Feature =
         |> Map.add (Feature.BasicMultiPartPayment) ([Feature.PaymentSecret])
         |> Map.add (Feature.PaymentSecret) ([Feature.VariableLengthOnion])
         
-    let private isFeatureOn(features: BitArray) (bit: int) =
+    let isFeatureOn(features: BitArray) (bit: int) =
         (features.Length > bit) && features.Reverse().[bit]
         
     let hasFeature(features: BitArray) (f: Feature) (support: FeaturesSupport option) =
@@ -170,17 +175,14 @@ module internal Feature =
         |> Set
         
         
-/// Uses regular class instead of F# type for caching byte[] representation
 [<StructuredFormatDisplay("{PrettyPrint}")>]
 type FeatureBit private (bitArray) =
-    let mutable bytes = null
     member val BitArray: BitArray = bitArray with get, set
     member this.ByteArray
         with get() =
-            if isNull bytes then
-                bytes <- this.BitArray.ToByteArray()
-            bytes
-        and set(v: byte[]) = bytes <- v
+            this.BitArray.ToByteArray()
+        and set(bytes: byte[]) =
+            this.BitArray <- BitArray.FromBytes(bytes)
     static member TryCreate(ba: BitArray) =
         result {
             do! Feature.validateFeatureGraph(ba)
@@ -196,12 +198,8 @@ type FeatureBit private (bitArray) =
         let b: bool array = [||]
         b |> BitArray |> FeatureBit
     static member TryCreate(bytes: byte[]) =
-        result {
-            let! fb = FeatureBit.TryCreate(BitArray.FromBytes(bytes))
-            fb.ByteArray <- bytes
-            return fb
-        }
-        
+        FeatureBit.TryCreate(BitArray.FromBytes(bytes))
+
     static member TryCreate(v: int64) =
         BitArray.FromInt64(v) |> FeatureBit.TryCreate
         
@@ -230,6 +228,27 @@ type FeatureBit private (bitArray) =
     override this.ToString() =
         this.BitArray.PrintBits()
         
+    member this.SetFeature(feature: Feature) (support: FeaturesSupport) (on: bool): unit =
+        let index = feature.BitPosition support
+        let length = this.BitArray.Length
+        if length <= index then
+            this.BitArray.Length <- index + 1
+
+            //this.BitArray.RightShift(index - length + 1)
+
+            // NOTE: Calling RightShift gives me:
+            // "The field, constructor or member 'RightShift' is not defined."
+            // So I just re-implement it here
+            for i in (length - 1) .. -1 .. 0 do
+                this.BitArray.[i + index - length + 1] <- this.BitArray.[i]
+
+            // NOTE: this probably wouldn't be necessary if we were using
+            // RightShift, but the dotnet docs don't actualy specify that
+            // RightShift sets the leading bits to zero.
+            for i in 0 .. (index - length) do
+                this.BitArray.[i] <- false
+        this.BitArray.[this.BitArray.Length - index - 1] <- on
+
     member this.HasFeature(f, ?featureType) =
         Feature.hasFeature this.BitArray (f) (featureType)
         
@@ -249,12 +268,8 @@ type FeatureBit private (bitArray) =
         
     // --- equality and comparison members ----
     member this.Equals(o: FeatureBit) =
-        if this.BitArray.Length <> o.BitArray.Length then false else
-        let mutable result = true
-        for i in 0..this.BitArray.Length - 1 do
-            if this.BitArray.[i] <> o.BitArray.[i] then
-                result <- false
-        result
+        this.ByteArray = o.ByteArray
+
     interface IEquatable<FeatureBit> with
         member this.Equals(o: FeatureBit) = this.Equals(o)
     override this.Equals(other: obj) =
--- a/src/DotNetLightning.Core/Utils/Extensions.fs
+++ b/src/DotNetLightning.Core/Utils/Extensions.fs
@@ -101,11 +101,34 @@ type BitArrayExtensions() =
 type System.Collections.BitArray with
     member this.ToByteArray() =
         if this.Length = 0 then [||] else
-        let ret: byte[] = Array.zeroCreate (((this.Length - 1) / 8) + 1)
-        let boolArray: bool[] = this.Reverse()
-        let t = BitArray(boolArray)
-        t.CopyTo(ret, 0)
-        ret |> Array.rev
+
+        let leadingZeros =
+            match (Seq.tryFindIndex (fun b -> b) (Seq.cast this)) with
+            | Some i -> i
+            | None -> this.Length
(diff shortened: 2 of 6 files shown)


src/DotNetLightning.Core/Utils/Extensions.fs
```diff
@@ -148,11 +148,10 @@ type System.Collections.BitArray with
         BitArray.From5BitEncoding(b)
         
     member this.Reverse() =
-        let length = this.Length
-        let mutable result = Array.zeroCreate length
-        for i in 0 .. (length - 1) do
-            result.[i] <- this.[length - i - 1]
-        result
+        let boolArray: array<bool> = Array.ofSeq (Seq.cast this)
+        Array.Reverse boolArray
```
knocte, line 152: there's no need to hardcode `Array` here, `.Reverse` exists in System.Linq and can be applied over IEnumerable

knocte:
@canndrew the description of the PR is very long and descriptive, cool! why don't you paste it in the commit message of the first commit?

src/DotNetLightning.Core/Serialize/Features.fs
```diff
@@ -230,6 +224,29 @@ type FeatureBit private (bitArray) =
...
         
+    member this.SetFeatureBit(index: int, on: bool): unit =
+        let length = this.BitArray.Length
+        if length <= index then
+            this.BitArray.Length <- index + 1
+
+            //this.BitArray.RightShift(index - length + 1)
+
+            // NOTE: Calling RightShift gives me:
+            // "The field, constructor or member 'RightShift' is not defined."
```
joemphilips, line 235: Probably because we are targeting netstandard2.0 instead of 2.1.
I forgot why we are still using 2.0.

Defining by ourselves is fine but ideally you can 
1. change `<TargetFramework>netstandard2.0` to `<TargetFrameworks>netstandard2.0;netstandard2.1`
1. use compilation flag to use RightShift only when the target is netstandard2.1

src/DotNetLightning.Core/Serialize/Features.fs
```diff
@@ -230,6 +224,29 @@ type FeatureBit private (bitArray) =
     override this.ToString() =
         this.BitArray.PrintBits()
         
+    member this.SetFeatureBit(index: int, on: bool): unit =
```
joemphilips, line 227: Why not take `Feature` type as an argument? isn't that safer because the API consumer cannot pass a wrong index number?

joemphilips: I mean both `Feature` and `FeaturesSupport` as an argument.

src/DotNetLightning.Core/Serialize/Features.fs
```diff
@@ -173,14 +173,12 @@ module internal Feature =
 /// Uses regular class instead of F# type for caching byte[] representation
```
joemphilips, line 173: This comment should be deleted, I suppose.

src/DotNetLightning.Core/Serialize/Features.fs
```diff
@@ -230,6 +224,29 @@ type FeatureBit private (bitArray) =
...
+                this.BitArray.[i + index - length + 1] <- this.BitArray.[i]
+
+            // NOTE: this probably wouldn't be necessary if we were using
+            // RightShift, but the dotnet docs don't actualy specify that
+            // RightShift sets the leading bits to zero.
+            for i in 0 .. (index - length) do
+                this.BitArray.[i] <- false
+        this.BitArray.[this.BitArray.Length - index - 1] <- on
+
+    member this.GetFeatureBit(index: int): bool =
```
joemphilips, line 247: Also, I suppose this should take `Feature` and `FeaturesSupport` too, and `IsFeatureOn` sounds better name. because it is more obvious that the return value is bool.

joemphilips: And I think this method should be named `HasFeature` too,  because it is more obvious that the return value is bool.

joemphilips: or `IsFeatureOn` you can choose.

src/DotNetLightning.Core/Utils/Extensions.fs
```diff
@@ -101,11 +101,34 @@ type BitArrayExtensions() =
 type System.Collections.BitArray with
     member this.ToByteArray() =
```
joemphilips, line 102: The method has become quite lengthy, can you add a test to assure that it is actually the inverse of `FromBytes` somehow?
It can be either property based test or regular unit test. And you can test the behavior via `FeatureBit` instead of testing this method directly if that sounds better.

src/DotNetLightning.Core/Utils/Extensions.fs
canndrew: Unfortunately `BitArray` can't be constructed from a sequence (as far as I could tell), so I still need to explicitly convert to an `Array`. So whether I call `Array.Reverse` or `Linq.Enumerable.Reverse` is kinda inconsequential.

src/DotNetLightning.Core/Serialize/Features.fs
canndrew: I'm not sure that's a good idea. If we use a compilation flag then we'll still need to implement `RightShift` ourselves for when we're using 2.0. So it wouldn't help reduce bugs or make the code smaller or simpler. It would just make it inconsistent between two slightly different build configurations.

src/DotNetLightning.Core/Serialize/Features.fs
canndrew: Okay, I've changed it.

src/DotNetLightning.Core/Serialize/Features.fs
canndrew: Done.

src/DotNetLightning.Core/Serialize/Features.fs
canndrew: I've removed `GetFeature` and just use `HasFeature` instead.

src/DotNetLightning.Core/Utils/Extensions.fs
canndrew: Okay, I've modified the property test that was there and added an extra unit test.

canndrew:
> @canndrew the description of the PR is very long and descriptive, cool! why don't you paste it in the commit message of the first commit?

Done.

tests/DotNetLightning.Core.Tests/Serialization.fs
```diff
@@ -742,8 +742,20 @@ module SerializationTest =
                 ()
                 
             testProperty "BitArray serialization" <| fun (ba : NonNull<byte[]>) ->
-                Expect.sequenceEqual (BitArray.FromBytes(ba.Get).ToByteArray()) (ba.Get) ""
-                
+                let backAndForth = BitArray.FromBytes(ba.Get).ToByteArray()
+                let finalArray = Array.zeroCreate ba.Get.Length
+                Array.Copy(backAndForth, 0, finalArray, ba.Get.Length - backAndForth.Length, backAndForth.Length)
+                Expect.isTrue false "tits"
```
joemphilips, line 748: WTF

joemphilips: Oh gosh some tests are not running  on CI.

joemphilips: I made a PR to fix this #93 .
I want you to rebase this branch on master after merging #93 . @canndrew

canndrew: Oops :grimacing:

That was some dummy code to check that the test was actually running. I forgot to take it out.

canndrew:
I've removed the dummy code and rebased onto master.

knocte:
@canndrew core tests are failing

knocte:
Still failing 😬

canndrew:
Hopefully fixed now...

knocte:
Nope :(

canndrew:
Okay, I've finally fixed it.

What was happening is that the payment request encoding converts a feature bit array to bech32 format by first padding zero bits until it has a multiple of 5 bits, then converting to a byte array, then converting the byte array to bech32. The problem was that `ToByteArray` now drops redundant 0 bits from the front of the feature bit array, meaning that the padding got removed so the resulting byte array didn't have a multiple of 5 bits in it, and when `InternalBech32Encoder` converts a byte array to bech32 it will pad zero bits to the *end* of the array to bring it up to a multiple of 5 bits.

In this case `InternalBech32Encoder.ConvertBits` should probably throw an error rather than pad, which it can do if passed `false` for its `pad` argument. But the call to `InternalBech32Encoder.ConvertBits` is wrapped in a wrapper which always passes `true`.

I've fixed it by first converting the bit array to a byte array, then padding the byte array, rather than the other way around.

knocte:
@joemphilips ready for review

canndrew:
I think the lesson here is that proper typing is important. Having the `ToByteArray` method on `BitArray` doesn't quite make sense since a `BitArray` is just an arbitrary array of bits, and not necessarily a multiple of 8 bits long. It's only by becoming a `FeatureBit` that the zeros at the front of the array become meaningless, which allows them to be trimmed or padded as necessary when we want to convert to bytes.

Both `ToByteArray` and `FromBytes` also flip the order of bits within bytes, which again only really makes sense given how we happen to be using them. If `BitArray.FromBytes` was in the standard library rather than an extension method it would behave differently - by having `bit.[0]` become the least-significant bit of the first byte of the resulting byte array and so forth, ie. the same behaviour that the built-in `BitArray.CopyTo` method has.

Since the extension methods on `BitArray` were already specially-crafted for their use in `FeatureBit` I didn't see the harm in modifying `BitArray.ToByteArray` to drop the unused bytes (this also fixed a bug in init message serialization in the process). But it turns out another piece of the code was relying on those zero bytes not being dropped. This is what happens when a method doesn't have clearly-defined intended semantics :/

joemphilips:
@canndrew Nice summary.
Yeah it was my fault that I did not consider bech32 encoding when I started to write `FeatureBit`. So it ended up in an awkward way.

Test looks good. Going to merge.

PR #93, opened 2020-05-07
use dotnet run to run expecto tests

joemphilips:
Running expecto with `dotnet test` worked before, but it has been ignoring all tests when we run with `dotnet test`

This is caused by version mismatch between `YoloDev.Expecto.TestSdk` and `Expecto` itself, When I change Expecto version to `"*"` , it worked fine.
We were using version `8.*` of expecto, and YoloDev.Expecto.TestSdk were using Expecto version `9.*` internally... And that caused this behavior.

There are many solutions to this, but I guess I should instead just run tests by `dotnet run` , that sounds more simpler in many ways.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,16 +19,24 @@ jobs:
       # we want to run only once.
       if: startsWith(matrix.os, 'ubuntu-18')
       run: |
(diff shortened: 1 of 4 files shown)


joemphilips:
I also reverted 8a761ff7fedb0c20615762d85d8713cd49c25d7a in #88 since that was causing test failure.

PR #94, opened 2020-05-13
Fix bug in accept_channel msg validation

canndrew:
The function `checkDustLimitIsLargerThanOurChannelReserve` was calculating the channel reserve based on the funding amount of the` open_channel` message, rather than using the channel reserve that was actually sent in the `open_channel` message.

This fixes the bug.


joemphilips:
Yep you are right. Ideally there should be a test for these validations but it is too tedious to write it.

PR #95, opened 2020-05-20
Use MaxFeeRateMismatchRatio in checkRemoteFee

canndrew:
The `OpenChannelMsgValidation.checkRemoteFee` method rejects any `open_channel` message with a fee rate less than the estimated fee rate. This PR changes it to use the provided `MaxFeeRateMismatchRatio` config option.

### DIFF
--- a/src/DotNetLightning.Core/Channel/ChannelError.fs
+++ b/src/DotNetLightning.Core/Channel/ChannelError.fs
@@ -193,6 +193,11 @@ module private ValidationHelper =
 /// Helpers to create channel error
 [<AutoOpen>]
 module internal ChannelError =
+    let feeRateMismatch (FeeRatePerKw remote, FeeRatePerKw local) =
+        let remote = float remote
+        let local = float local
+        abs (2.0 * (remote - local) / (remote + local))
+
     let inline feeDeltaTooHigh msg (actualDelta, maxAccepted) =
         InvalidUpdateFeeError.Create
             msg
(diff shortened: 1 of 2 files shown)


canndrew:
One thing I noticed when doing this is that the `MaxFeeRateMismatchRatio` option is used weirdly. Given that it has "ratio" in the name, I would expect a value of 1.0 to mean that the fee rates must match exactly, a fee rate of 0.5 or 2.0 to mean that the counterparty's fee rate must be between 50% to 200% of ours, and so forth. In the code though, the formula for calculating the "ratio" is:

    let feeRateMismatch (FeeRatePerKw remote, FeeRatePerKw local) =
        (2.0 * float (remote - local) / float (remote + local))
        |> abs

Which means a value of 1.0 actually corresponds to the fee rates being allowed to differ by 33-300 percent, a value of 0.0 means they must match, etc.

Should this config option be re-named or should this formula be modified?

knocte:
IMO that ratio should not be a primitive type, but a type which has some validation in its constructor (so that it cannot be negative and it cannot be higher than MAX).

After doing that, IMO a more readable approach is that MAX=100.0, not 1.0.

src/DotNetLightning.Core/Channel/ChannelError.fs
```diff
@@ -193,6 +193,10 @@ module private ValidationHelper =
 /// Helpers to create channel error
 [<AutoOpen>]
 module internal ChannelError =
+    let feeRateMismatch (FeeRatePerKw remote, FeeRatePerKw local) =
+        (2.0 * float (remote - local) / float (remote + local))
```
joemphilips, line 197: @knocte 's comment made me realize that this should actually be
```
(2.0 * ((float remote) - (float local ) / float (remote + local) ))
```

joemphilips:
hmm, for the `feeRateMismatch` , I did not think much about it when I was porting from the [original function in an eclair](https://github.com/ACINQ/eclair/blob/029cafe297dba049b76faa2344de0f4d63aa97d0/eclair-core/src/main/scala/fr/acinq/eclair/channel/Helpers.scala#L195)

> After doing that, IMO a more readable approach is that MAX=100.0, not 1.0.

There are other place in the code that we use `float` as a ratio, and it all uses 1.0 as a 100% .
So lets keep it that way to make it consistant.

knocte:
> There are other place in the code that we use float as a ratio, and it all uses 1.0 as a 100% .
> So lets keep it that way to make it consistant.

Ok agreed. Then @canndrew do something like this:

```
type Ratio(ratio: float) =
    do
        if ratio < 0 || ratio > 1 then
            invalidArg "..."
```

src/DotNetLightning.Core/Channel/ChannelError.fs
joemphilips: I mean
```fs
(2.0 * (float remote) - (float origin)) /
(float (remote + local))
```

canndrew:
To me, "ratio" conveys a rational number, and not necessarily in the range (0, 1). Also, there's no reason why the `MaxFeeRateMismatchRatio` in this case can't be greater than 1. It would be silly to set it that high, but it would also be silly to set it higher than about 0.5. There's nothing inherently special about the value `1` in that formula.

src/DotNetLightning.Core/Channel/ChannelError.fs
canndrew: I've fixed this.

joemphilips:
The reason for multiplying by 2.0 is still in mystery, but let's just merge this and think about it later when we find a pain.

knocte:
> The reason for multiplying by 2.0 is still in mystery

@canndrew please open an issue in eclair repo to clarify this

PR #96, opened 2020-05-20
Fix checkIfFundersAmountSufficient

canndrew:
OpenChannelMsgValidation.checkIfFundersAmountSufficient incorrectly only fails if both the channel amounts are less than their channel reserves, rather than if either of them are.

This fixes the bug.


canndrew:
I just noticed that this function isn't actually called from anywhere, though it seems that it should be called from the channel validation module. Is this just an oversight?

joemphilips:
Ah you are right, it should be
```fs
        *^> OpenChannelMsgValidation.checkIfFundersAmountSufficient (feeEstimator) msg
```

in `ChannelValidation.fs` line 162. I've forgot to call.

canndrew:
Actually this function is broken as-is. It checks whether `push_msat` is larger than the channel reserve but this shouldn't be expected to be the case. The counterparty is allowed to have less than `channel_reserve` funds in the channel to start with, they're just not allowed to make payments that bring them below the channel reserve.

canndrew:
I'm going to make another PR that splits this function into two and removes the `toLocalMSat <= (msg.ChannelReserveSatoshis.ToLNMoney())` check.

canndrew:
This PR has been superseded by #97.

PR #97, opened 2020-05-20
Outgoing open channel validation

canndrew:
This builds off of #95 and can be rebased if any further changes occur on that PR.

This PR adds validation to the `CreateOutbound` channel command which enforces that the channel parameters are sensible - ie. that the channel wouldn't be rejected by any sane and spec-compliant peer.

It also removes the (unused) `checkIfFundersAmountSufficient` function and moves one of it's checks into a new `checkFunderCanAffordFee` function. The other two checks were unnecessary and overly restrictive. One of them checked that `push_msat` is greater than the fundee's `channel_reserve`, which is unlikely to ever be the case. The other checked that the funder's initial channel balance minus their channel reserve is enough to cover commitment tx fees. This is also unnecessary since a channel is closed when a commitment tx is broadcast and so the channel reserve is no longer relevant. Simply checking that the funder's initial balance is enough to cover the fee is sufficient.

### DIFF
--- a/src/DotNetLightning.Core/Channel/ChannelValidation.fs
+++ b/src/DotNetLightning.Core/Channel/ChannelValidation.fs
@@ -146,9 +146,17 @@ module internal ChannelHelpers =
 module internal Validation =
 
     open DotNetLightning.Channel
-
+    let checkOurOpenChannelMsgAcceptable (conf: ChannelConfig) (msg: OpenChannel) =
+        Validation.ofResult(OpenChannelMsgValidation.checkFundingSatoshisLessThanMax msg)
+        *^> OpenChannelMsgValidation.checkChannelReserveSatohisLessThanFundingSatoshis msg
+        *^> OpenChannelMsgValidation.checkPushMSatLesserThanFundingValue msg
+        *^> OpenChannelMsgValidation.checkFundingSatoshisLessThanDustLimitSatoshis msg
+        *^> OpenChannelMsgValidation.checkMaxAcceptedHTLCs msg
+        *^> OpenChannelMsgValidation.checkFunderCanAffordFee (msg.FeeRatePerKw) msg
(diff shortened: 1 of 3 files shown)


src/DotNetLightning.Core/Channel/ChannelValidation.fs
```diff
@@ -146,19 +146,28 @@ module internal ChannelHelpers =
...
     open DotNetLightning.Channel
-
+    let checkOurOpenChannelMsgAcceptable (conf: ChannelConfig) (msg: OpenChannel) =
+        Validation.ofResult(OpenChannelMsgValidation.checkFundingSatoshisLessThanMax msg)
+        *^> OpenChannelMsgValidation.checkChannelReserveSatohisLessThanFundingSatoshis msg
+        *^> OpenChannelMsgValidation.checkPushMSatLesserThanFundingValue msg
+        *^> OpenChannelMsgValidation.checkFundingSatoshisLessThanDustLimitSatoshis msg
+        *^> OpenChannelMsgValidation.checkMaxAcceptedHTLCs msg
+        *^> OpenChannelMsgValidation.checkFunderCanAffordFee (msg.FeeRatePerKw) msg
+        |> Result.mapError(InvalidOpenChannelError.Create msg >> InvalidOpenChannel)
```
joemphilips, line 156: It may be good to add some message so that we can distinguish an error returned by `checkOurOpenChannelMsgAcceptable` and by `checkOpenChannelMsgAcceptable`
e.g. change this line to
```fs
       |> Result.mapError((@)["our open_channel msg is invalid"] >> InvalidAcceptChannelError.Create msg >> InvalidAcceptChannel)
```

canndrew: Done. I've added a similar message for their open_channel message too.

knocte:
@joemphilips why did CI fail in the commit that merged this PR?

joemphilips:
@knocte Infrastructure tests arebit flaky now. I will search what is causing this.

PR #101, opened 2020-05-27
add package for macaroon and its test

joemphilips:
Based on https://github.com/JornWildt/Macaroons.Net

Strictly speaking this is an API layer thing. So I wasn't sure I should include it in here.
But it seems people in lightning lab are trying to standardize its use for [LN-powered authentication](https://lightning.engineering/posts/2020-03-30-lsat/)

So I will put in here. Probably I will use this package from `DotNetLightning.Core` to expose LSAT-spec compliant APIs.


knocte:
Do you mind synchronizing commit message with PR description? otherwise if we ever move DNL to other location we lose that info.

joemphilips:
Finished. I will merge this in a few days.

PR #102, opened 2020-05-27
Add Message member to error types

canndrew:
Add a `Message` member to error types which returns an error message suitable for displaying to the user, in contrast with `ToString` which returns a debugging-friendly textual representation of the error object.

This MR also expands on and attempts to improve some of the existing error messages.

### DIFF
--- a/src/DotNetLightning.Core/Channel/ChannelError.fs
+++ b/src/DotNetLightning.Core/Channel/ChannelError.fs
@@ -79,10 +79,10 @@ type ChannelError =
         | InvalidOperationAddHTLC _ -> Ignore
         | RemoteProposedHigherFeeThanBefore(_, _) -> Close
     
-    override this.ToString() =
+    member this.Message =
         match this with
         | HTLCAlreadySent htlcId ->
-            sprintf "We have already sent a fail/fulfill for this htlc: %A" htlcId
+            sprintf "We have already sent a fail/fulfill for htlc id %i" htlcId.Value
         | InvalidPaymentPreimage(e, actual) ->
             sprintf "Invalid HTLC PreImage %A. Hash (%A) does not match the one expected %A"
                     actual
@@ -98,8 +98,8 @@ type ChannelError =
             sprintf "Number of signatures went from the remote (%A) does not match the number expected (%A)" actual expected
         | TheyCannotAffordFee (toRemote, fee, channelReserve) ->
             sprintf "they are funder but cannot afford their fee. to_remote output is: %A; actual fee is %A; channel_reserve_satoshis is: %A" toRemote fee channelReserve
-        | InvalidOpenChannel x ->
-            sprintf "Invalid open_channel from the peer. \n %s" (x.ToString())
+        | InvalidOpenChannel invalidOpenChannelError ->
+            sprintf "Invalid open_channel from the peer.: %s" invalidOpenChannelError.Message
         | OnceConfirmedFundingTxHasBecomeUnconfirmed (height, depth) ->
             sprintf "once confirmed funding tx has become less confirmed than threshold %A! This is probably caused by reorg. current depth is: %A " height depth
         | ReceivedShutdownWhenRemoteHasUnsignedOutgoingHTLCs msg ->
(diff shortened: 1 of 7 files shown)


knocte:
Telling anything to the user about HTLC doesn't seem to be user-friendly at all.

src/DotNetLightning.Core/Channel/ChannelError.fs
```diff
@@ -79,10 +79,10 @@ type ChannelError =
         | InvalidOperationAddHTLC _ -> Ignore
         | RemoteProposedHigherFeeThanBefore(_, _) -> Close
     
-    override this.ToString() =
+    member this.Message =
```
knocte, line 82: IMO you should not remove ToString(), that's a valid way for developers to find a DEBUG-representation of something (so it's fine if it's JSON)

canndrew: This will still have a `ToString` method. It will just be the default one which prints the contents of the object, which is what you want for debugging.

canndrew:
> Telling anything to the user about HTLC doesn't seem to be user-friendly at all.

The only error messages that mention HTLCs are ones that indicate protocol violations by one or other of the parties, not ones that users should normally see. If a user does see one then giving them the details of the error seems strictly better than not giving it to them, and it allows them to file a useful bug report.

src/DotNetLightning.Core/Channel/ChannelError.fs
joemphilips: I don't see the point of having two kinds of string representation.
`ToString` should serve for both debugging and for showing the message to the user.

canndrew: .NET's exceptions have distinct `ToString` and `Message` members. The point is that `Message` gives a user-friendly string and `ToString` dumps a textual representation of the entire error object, including the exact value of all the fields.

What prompted this change is that geewallet was printing error message that looked like this:

```
Error opening channel: Peer responded to our open_channel with an error message: Invalid open_channel from the peer. 
 { Msg =
       { Chainhash =
                    6fe28c0ab6f1b372c1a6a246ae63f74f931e8365e15a089c68d6190000000000
         TemporaryChannelId =
                             ChannelId
                               05a54268153a61c35ded693133aaeeb3c3ac12903e4e4dac215f05aeb5ef9a05
         FundingSatoshis = 0.00030000
         PushMSat = LNMoney 0L
         DustLimitSatoshis = 0.00000005
         MaxHTLCValueInFlightMsat = LNMoney 5000L
         ChannelReserveSatoshis = 0.00000300
         HTLCMinimumMsat = LNMoney 1000L
         FeeRatePerKw = FeeRatePerKw 16357u
         ToSelfDelay = BlockHeightOffset16 6us
         MaxAcceptedHTLCs = 10us
         FundingPubKey =
                        02d39a4e5a41ab40d158684daf16651de2f87a8a8b8e0bf4167a99da52b3be2272
         RevocationBasepoint =
                              03c21ad70f41bb9fc6efa36853c88266c0ca0d48ee61734f4e86380d5e5b46ebe0
         PaymentBasepoint =
                           0249aee64aab597d30e80e11ae4464f9a1b3e2beefdcd2be7535a41b6c4173cf6a
         DelayedPaymentBasepoint =
                                  03d987ff1a55a5a8aae8368970f2d9fa6d2d71a33b4c2441dab60b33c37eff979c
         HTLCBasepoint =
                        02760a472b45b1315db70931378c727367147520a7938389048b021a7c0f072973
         FirstPerCommitmentPoint =
                                  027dfbdea98d8271f924b64b698af2f0995362afe47454c76e70bb7254e78b9ecd
         ChannelFlags = 0uy
         ShutdownScriptPubKey =
                               Some
                                 OP_HASH160 2f24ec0dfc2a1beddb3619f38b62737b19e28d6c OP_EQUAL }
  Errors =
          ["their open_channel msg is invalid";
           "Suitable channel reserve not found. Aborting. (our channel reserve was (0.00000300). and our dust limit was(0.00001701))";
           "channel_reserve_satoshis too small. It was: 0.00000300; but our dust_limit is: 0.00001701"] }
```

Dumping the `Msg` field like that isn't really good UX. Plus the error that's being printed there even has an `Errors` field consisting of a list of strings that look intended to be displayed to the user.

The error type in this case is `InvalidOpenChannelError`. With this PR that type still has its built-in `ToString` method which prints all of the above and is useful for developers, but it also has a `Message` method which returns a concatenation of the `Errors` strings (separated by `; `). So geewallet just calls `Message` when it displays the error to the user.

PR #103, opened 2020-05-28
Avoid newer F# syntax

ysangkok:
This allows for compiling on Mono, which still ships with F# 4.5.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: Build DotNetLightning and deploy to NuGet
+name: Build and deploy
 on: [push, pull_request]
 jobs:
   build_and_test:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,3 +31,15 @@ jobs:
     - name: Run other tests
       run: |
         dotnet test
+
+  build_with_fsharp_from_mono:
+    runs-on: ubuntu-18.04
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build
+      run: |
+        DEBIAN_FRONTEND=noninteractive sudo apt install -y msbuild fsharp nuget
+
+        nuget restore DotNetLightning.sln
+        msbuild src/DotNetLightning.Core/DotNetLightning.Core.fsproj -p:BouncyCastle=True -p:TargetFramework=netstandard2.0
+
--- a/src/DotNetLightning.Core/Payment/PaymentRequest.fs
+++ b/src/DotNetLightning.Core/Payment/PaymentRequest.fs
@@ -176,16 +176,16 @@ type FallbackAddress = private {
     member this.ToAddress(prefix: string) =
         match this.Version with
         | 17uy when prefix = "lnbc" ->
-            let data = Array.concat (seq { [| Helpers.PREFIX_ADDRESS_PUBKEYHASH |]; this.Data })
+            let data = Array.concat (seq { yield [| Helpers.PREFIX_ADDRESS_PUBKEYHASH |]; yield this.Data })
             Helpers.base58check.EncodeData(data)
         | 18uy when prefix = "lnbc" ->
-            let data = Array.concat (seq { [| Helpers.PREFIX_ADDRESS_SCRIPTHASH |]; this.Data })
+            let data = Array.concat (seq { yield [| Helpers.PREFIX_ADDRESS_SCRIPTHASH |]; yield this.Data })
             Helpers.base58check.EncodeData(data)
         | 17uy when prefix = "lntb" || prefix = "lnbcrt" ->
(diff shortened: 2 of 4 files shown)


src/DotNetLightning.Core/Payment/PaymentRequest.fs
```diff
@@ -79,12 +79,12 @@ module private Helpers =
     let prefixValues = prefixes |> Map.toList |> List.map(fun (_, v) -> v)
         
     let checkAndGetPrefixFromHrp (hrp: string) =
-        let maybePrefix = prefixValues |> List.filter(fun p -> hrp.StartsWith(p)) |> List.tryExactlyOne
+        let maybePrefix = prefixValues |> List.filter(fun p -> hrp.StartsWith(p))
```
joemphilips, line 82: You should add polyfill for `tryExactlyOne` in Utils/Extensions.fs` , in that way the change will be much simpler.

knocte:
@ysangkok why do you need this? we consume DNL's .netstandard lib which is generated by .NETCore.

src/DotNetLightning.Core/Payment/PaymentRequest.fs
knocte: or use System.Linq's `.Single()`

ysangkok:
Since geewallet is not a net core project, it is more convenient to use
dotnetlightning as a project reference and build it with msbuild. But then
i am using monos fsharp which is 4.5

knocte:
@joemphilips rebased and simplified PR (diff is much shorter now).

Also added new CI job to:
- demonstrate that the simplification still compiles.
- avoid breaking F# 4.5 compatibility by mistake, moving forward.

knocte:
PS: Better not to squash this when merging.

.github/workflows/ci.yml
```diff
@@ -31,3 +31,18 @@ jobs:
...
+  build_with_fsharp_from_mono:
+    runs-on: ubuntu-18.04
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build
+      run: |
+        DEBIAN_FRONTEND=noninteractive sudo apt install -y msbuild fsharp nuget
+
+        # override TargetFrameworks to not build netstandard2.1 (incompatible with Mono)
+        sed -i "7i<PropertyGroup><TargetFrameworks>netstandard2.0</TargetFrameworks></PropertyGroup>" src/DotNetLightning.Core/DotNetLightning.Core.fsproj
```
joemphilips, line 44: Is this necessary?  Is it impossible to specify the target framework from the command line argument for msbuild?

.github/workflows/ci.yml
```diff
@@ -31,3 +31,18 @@ jobs:
...
       run: |
         dotnet test
+
+  build_with_fsharp_from_mono:
+    runs-on: ubuntu-18.04
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build
+      run: |
+        DEBIAN_FRONTEND=noninteractive sudo apt install -y msbuild fsharp nuget
```
joemphilips, line 41: Would you specify the version for these packages? I think you can pass it to apt command by `apt install -y msbuild=<version>`

.github/workflows/ci.yml
knocte: I guess not, but I'll double check

.github/workflows/ci.yml
knocte: why would you want to do this?

.github/workflows/ci.yml
knocte: you were right; fixed

.github/workflows/ci.yml
joemphilips: Because the purpose of this job is to demonstrate that it can be compiled on a specific fsharp version? The "latest" version  might be changed by runtime and it may be difficult to reproduce when the build is broken.
(TBH I don't have strong opinion about this though. I can merge as is if you feel not to specify.)

knocte: Ah I suspected that you might be thinking along those lines. However, the point of this compatibility is that DNL.Core stays compatible to the last version of fsharp's apt package provided by Mono's apt sources. If they upgrade fsharp package to a higher version, then that's not a bad thing; I mean, then this CI would not be checking anymore against v4.5 but against the last version they bundle, but in case it lags behind (again) compared to the last version of F# provided by .NETCore, then the CI becomes useful again.

On top of that, specifying a version of the package might even not work at all. Don't quote me on this but I think that, whenever an apt source is upgraded with a new version of a package, the old version doesn't exist anymore (and cannot be retrieved). Maybe the only useful thing that comes of specifying a version when doing an apt installation is in case there are two different apt sources enabled, which provide the same package but with different version.

(That being said, there is actually an fsharp package with version older than 4.5 -- IIRC it's 4.0 -- which comes from Canonical repos, as opposed to Mono's repos. However, IME this version is not compatible with msbuild.)

joemphilips: I see.

PR #107, opened 2020-06-09
Add `Msg` suffix to message type names

canndrew:
This makes code that handles messages more comprehensible.


src/DotNetLightning.Core/Channel/Channel.fs
```diff
@@ -255,22 +258,23 @@ module Channel =
...
-                                      MaxAcceptedHTLCs = localParams.MaxAcceptedHTLCs
-                                      FundingPubKey = channelKeys.FundingKey.PubKey
-                                      RevocationBasepoint = channelKeys.RevocationBaseKey.PubKey
-                                      PaymentBasepoint = channelKeys.PaymentBaseKey.PubKey
-                                      DelayedPaymentBasepoint = channelKeys.DelayedPaymentBaseKey.PubKey
-                                      HTLCBasepoint = channelKeys.HTLCBaseKey.PubKey
-                                      FirstPerCommitmentPoint = localCommitmentSecret.PubKey
-                                      ShutdownScriptPubKey = cs.Config.ChannelOptions.ShutdownScriptPubKey }
-
+                let acceptChannel: AcceptChannelMsg = {
```
knocte, line 261: I don't think you need to add types, the most important thing is renaming the variable to add the "Msg" suffix, which you forgot to do here

joemphilips: To me renaming sounds not necessary, it is not a big deal.

canndrew: I moved the type annotation

```
let acceptChannel = { AcceptChannel.TemporaryChannelId
                      ^^^^^^^^^^^^^
                                   - from here
```

```
let acceptChannel: AcceptChannelMsg {
                   ^^^^^^^^^^^^^^^^
                                   - to here
```

Because another similar line of code in the file required me to (because adding the suffix screwed up the indentation).

Adding the suffix to the variable name is a good idea too though. I'll edit the commit.

joemphilips:
To me naming sounds not a big deal. But if you think this is better, I'm not against it.
Should I merge now or you want to prefix each variable names with `msg` ? @canndrew

knocte:
> But if you think this is better

I, for one, think it's way better; speaking as a somewhat outsider to the internals of DNL.

canndrew:
I've prefixed `Msg` to some variable names. There's a lot of variables that aren't named after the message type so I didn't search through and find them all.

PR #108, opened 2020-06-09
Add and switch to new types for revocation

canndrew:
Add the `UInt48`, `RevocationKey`, `CommitmentPubKey`, `CommitmentSeed` and `RevocationSet` types.

A `RevocationKey` is a secret key that can be used to revoke a transaction. It's shared in the `revoke_and_ack` message and stored in the user's wallet in a `RevocationSet`. A `CommitmentPubKey` is the public key of a `RevocationKey`, also known as a "per commitment point". A `CommitmentSeed` is the master revocation key for a side of a channel and can be used to derive all the revocation keys for that side of that channel.

A `RevocationSet` stores a set of revocation keys in a compact form specified in BOLT #3 which allows earlier keys to be derived from later keys.


joemphilips:
Thanks, wow this looks significant improvement.
Give me a day, I want to understand what is going on...

tests/DotNetLightning.Core.Tests/RevocationSetTests.fs
```diff
@@ -0,0 +1,281 @@
+module RevocationSetTests
+
+open NBitcoin
+open Expecto
+open DotNetLightning.Utils
+open DotNetLightning.Crypto
+
+let Unwrap<'T, 'E>(result: FSharp.Core.Result<'T, 'E>) : 'T =
```
joemphilips, line 8: nit: there is a `Result.deref` in ResultUtils to do this.

src/DotNetLightning.Core/Utils/UInt48.fs
```diff
@@ -0,0 +1,85 @@
...
+        UInt64 = 1UL
+    }
+
+    static member MaxValue: UInt48 = {
+        UInt64 = UInt48.BitMask
+    }
+
+    static member FromUInt64(x: uint64): UInt48 =
+        if x > UInt48.BitMask then
+            failwith <| sprintf "value %i out of range for UInt48" x
```
joemphilips, line 25: nit: raising `ArgumentOutOfRangeException` is better.

src/DotNetLightning.Core/Utils/NBitcoinExtensions.fs
```diff
@@ -35,3 +37,24 @@ module NBitcoinExtensions =
 
         member this.GetTxId() =
             this.GetGlobalTransaction().GetTxId()
+
+    type Key with
+        static member BytesLength: int = 32
+
+        member this.ToByteArray(): array<byte> =
```
joemphilips, line 44: Why you need this? `Key` already has `ToBytes()` method.

src/DotNetLightning.Core/Utils/NBitcoinExtensions.fs
```diff
@@ -35,3 +37,24 @@ module NBitcoinExtensions =
...
+            let mutable bytes: array<byte> = Array.zeroCreate Key.BytesLength
+            use bytesStream = new MemoryStream(bytes)
+            let stream = NBitcoin.BitcoinStream(bytesStream, true)
+            this.ReadWrite stream
+            bytes
+
+    type PubKey with
+        static member BytesLength: int = 33
+
+        member this.ToByteArray(): array<byte> =
```
joemphilips, line 54: Same as above.

joemphilips:
There are few nits but overall it seems good.

canndrew:
Thanks. I've fixed the nits (I didn't know about `ToBytes` or `deref`).

PR #109, opened 2020-06-16
Bump NBitcoin version and use `NBitcoin.Secp256k1` instead of `Secp256k1.Net`

joemphilips:
* NBitcoin has added some overload for taking Span,ReadOnlySpan
  as an argument. FSharp compiler complains that it does not know
  which overload to use if we pass byte[] as before.
  So make it explicit that we want to pass byte[] in some function call.
* Stop using `ForceLowR` option in tests.
  Test must use RFC 6974 signer so we should specify forceLowR=false
  when signing txs. PSBT.ForceLowR property is depreated and we should
  use SigningOptions object in NBitcoin instead.
* Set nSequence to 0 for HTLC-Success/Timeout txs.
  It is defined in bolt3 that we must set it to 0. This has been done implicitly
  by TransactionBuilder before, but the behaviour has changed in NBitcoin side.
  So we must set it explicitly.
* add `new` keyword when instantiating `Key` type.
  Since Key type has become IDisposable, and compiler emits warning when we
  did not add `new` for IDisposable.


src/DotNetLightning.Core/Channel/Channel.fs
```diff
@@ -44,7 +44,7 @@ module Channel =
 
     let private hex = NBitcoin.DataEncoders.HexEncoder()
     let private ascii = System.Text.ASCIIEncoding.ASCII
-    let private dummyPrivKey = Key(hex.DecodeData("0101010101010101010101010101010101010101010101010101010101010101"))
+    let private dummyPrivKey = new Key(hex.DecodeData("0101010101010101010101010101010101010101010101010101010101010101"))
```
knocte, line 47: if Key is not disposable, you don't need the `new` keyword

joemphilips: It has become disposable after Nicolas prepared `NBitcoin.Secp256k1`

knocte:
Does `NBitcoin.Secp256k1` provide the same as `Secp256k1.Net`? If yes, the latter ref should be removed?

joemphilips:
> Does NBitcoin.Secp256k1 provide the same as Secp256k1.Net? If yes, the latter ref should be removed?

Yes I'm working on.

src/DotNetLightning.Core/Channel/Channel.fs
knocte: oh I see; I guess because of the interop with unmanaged code?

joemphilips: There is no unmanaged code in NBitcoin, do you mean `NBitcoin.Secp256k1`, then yes.
It is because inner ECPrivKey object in NBitcoin.Secp256k1 is disposable.
I don't know why exactly this had to be disposable. Probably because in original secp256k1 it was clearing memory after using privkey to prevent sensitive infomation leaking in memory.

knocte: > do you mean NBitcoin.Secp256k1, then yes.

Yeah I mean NBitcoin.Secp256k1. Mmm, then I wonder if we should create a version of NBitcoin.Secp256k1 that uses BouncyCastle instead of unmanaged interop, what do you think?

joemphilips: > then I wonder if we should create a version of NBitcoin.Secp256k1 that uses BouncyCastle instead of unmanaged interop, what do you think?

I didn't get what you mean. I think you are misunderstanding that `NBitcoin.Secp256k1` is Native bindings to unmanaged code?  It is not. It is a pure C# port of the original [secp256k1](https://github.com/bitcoin-core/secp256k1).

knocte: good to know, but then I don't understand why it's disposable now

joemphilips: Because  `NBitcoin` depends on `NBitcoin.Secp256k1` and `NBitcoin.Key` holds `NBitcoin.Secp256k1.ECPrivKey` object. And this `ECPrivKey` is `IDisposable` so it makes sense to dispose it when the wrapping `Key` is disposed.

So why this ECPrivKey should be disposable? The reason I've mentioned above is my guess.

> I don't know why exactly this had to be disposable. Probably because in original secp256k1 it was clearing memory after using privkey to prevent sensitive infomation leaking in memory.

Nicolas wanted to make `NBitcoin.Secp256k1` to behave as much as the same with original secp256k1, He imitated the behavior of its finalizer with `Dispose` method.

joemphilips:
Done.

It is possible to replace Secp256k1 operation in BouncyCastle build (It is quite easy, just replace `newSecp256k1` method to return new implementation). But it is not done in this PR.
Because `NBitcoin.Secp256k1` seems not totally mature yet (See e.g. [this issue](https://github.com/MetacoSA/NBitcoin/issues/871)) I wanted to leave a choice to use BouncyCastle for Secp256k1 operation.

This PR also does not change the behavior of transport layer cryptography. (ChaCha20 based things).


I have a question to @knocte and @canndrew before merging this.

1. This bumps DNL's target framework from `netstandard2.0` to `netstandard2.1` because `NBitcoin.secp256k1` does not  support 2.0. Is this acceptable for your need?
It is possible to have multiple target frameworks and to support 2.0 only in BouncyCastle build . But I don't want to do that because Compilation and release process will be slightly complicated.

joemphilips:
> I have a question to @knocte and @canndrew before merging this.

Nevermind, It was much easier than I expected to support both versions.
See my last commit.

joemphilips:
I will merge this in few days.

PR #113, opened 2020-06-25
Core: add "Network" prefix to "Msg" member of error types

knocte:
It was very ambiguous to have a "Msg" member in an error type that
actually doesn't contain the error message, so we rename this to
"NetworkMsg".


joemphilips:
Yeah it makes sense to me too.

PR #116, opened 2020-07-02
Make ClosingData CLIMutable

Bobface:


knocte:
@Bobface next time please don't forget to write the *why* (not just the *what*) in the commit message.

PR #117, opened 2020-07-09
Fix ObscuredCommitmentNumber generation

canndrew:
Obscured commitment numbers treat the commitment number as counting upwards from 0 rather than downwards from `UInt48.MaxValue`.

This hadn't shown up in testing so far since we were only testing against ourselves and so the error cancelled out on both sides.


joemphilips:
LGTM.

PR #118, opened 2020-07-11
Always hold bytes in uint256 as a little endian

joemphilips:
When user constructs PaymentRequest from PaymentPreimage,
they usually gets it by `PaymentPreimage.Hash` member. But before this commit,
they had to reverse the order of the hash before giving it to PaymentRequest
constructor(as TaggeedFields). This is not ergonomic, and it breaks the
purpose of the member in the first place.
This commit reverses the order when deriving the hash.

This commit also changes the behavior of GetRIPEMD160 method.
(Because we were already using `PaymentPreimage.Hash` for BOLT3 tests, and we
had to reverse the order again before putting it into the commitment script)


joemphilips:
@knocte Please be careful if you are already using `.Hash` member in `PaymentPreimage` . This PR will change the behavior.

knocte:
@canndrew can you review?

canndrew:
Why does the hash need to be reversed? Does it say somewhere in the spec that the hash is transmitted in little-endian?

joemphilips:
> Why does the hash need to be reversed? Does it say somewhere in the spec that the hash is transmitted in little-endian?

@canndrew  No, it is always transmitted as a big-endian. Problem is that how we hold an inner value for `PaymentHash` for serialization.

Problem of the current serialization is, it reads a PaymentHash in UpdateAddHTLCMsg with `lendian = true`, which is not correct. So it will *not* reverse the big-endian encoded `payment_hash` into a small endian.
And before this PR, `PaymentPreimage.Hash` worked in that way too.

But `PaymentRequest` does not work in that way. When we pass `PaymentHash` as a tagged field, it assumes `PaymentHash` holds its inner value with a little-endian (which is a correct usage of uint256), so if we pass `PaymentHash` which either originated from P2P message or created by `PaymentPreimage.Hash`  to `PaymentRequest`, it results to malformed invoice.

So there are two ways to fix this.
1. Change serialization logic in `PaymentRequest` , so that we always assume uint256 will holds its inner value as big-endian
2. Change the inner representation of `PaymentHash`, so that its inner uint256 will always hold the value with little-endian.

I’ve chosen a second approach. Because that is the correct way to use `uint256` . When we convert `uint256` to string or vice versa, it will reverse the endian. So holding it wtih small endian will show up the right PaymentHash when we convert it to json or debug printing.

My first commit was in fact incomplete. It does change the behavior of `PaymentPreimage.Hash` to the right direction, but it does not change the network serialization.
See my second and third commit, I hope it is now clear to you.

joemphilips:
Probably the best way to handle these problems is to stop using `uint256` at all...

canndrew:
> I’ve chosen a second approach. Because that is the correct way to use uint256

This makes sense. I just wish it wasn't so easy to get the wrong endianess out of a uint256, since I just made the same mistake in my code and had to debug it. Some good protection against this would be to hide the internal `uint256` of all types that contain one (such as `TxId`, `ChannelId`, etc.) and to give those types their own `{To,From}{String,Bytes}` methods.

canndrew:
This LGTM. The inconsistency of reversing the endianess in arbitrary places was the only thing that was bothering me, but that's been fixed.

knocte:
I imagine this PR's commits need to be squashed now (or when merging)?

joemphilips:
> I imagine this PR's commits need to be squashed now (or when merging)?

Not very sure which is better, but I think I should not. granular commits are better in most cases.

joemphilips:
> This makes sense. I just wish it wasn't so easy to get the wrong endianess out of a uint256, since I just made the same mistake in my code and had to debug it. Some good protection against this would be to hide the internal uint256 of all types that contain one (such as TxId, ChannelId, etc.) and to give those types their own {To,From}{String,Bytes} methods.

> A Sha256Hash type might be nice too as an intermediate layer, since that would also guarantee that the uint256 is non-zero. This is useful for constructing Keys since you can't reliably construct one from an arbitrary array of 32 bytes, but you can from a random array of 32 bytes.

Yes, these are good for both performance and ergonomics. I'm not against to the refactoring. But I will delay my work until more pain comes out.

knocte:
> granular commits are better in most cases.

I like granular commits, except when they mean that if I checkout a particular snapshot of the repo from that commit, there's a bug. And in this case your incremental commits were fixing bugs introduced in the 1st.

PR #119, opened 2020-07-15
Use correct channel id in WaitForFundingSigned

canndrew:
When we are waiting for the `funding_signed` message we already know the true channel id of the channel, rather than just the temporary channel id. As such, we can store the true channel id in `WaitForFundingSignedData`. This fixes a bug in geewallet where querying the channel id of a channel would return the temporary channel id rather than the true channel id.


joemphilips:
Thanks, LGTM.

joemphilips:
@canndrew  Would you check if you are able to merge this by yourself after I have approved?
If not I will rebase and merge.

canndrew:
I don't have permission to merge here. I've rebased it off of latest master though.

PR #120, opened 2020-07-16
Make ClosingData public instead of CLIMutable

Bobface:
The ClosingData type was CLIMutable to allow for json
serialization and deserialization.

However, that only avoided an exception being thrown.
The data was lost when writing and reading the json.

This commit removes the CLIMutable attribute and
makes the type public as the other types.


joemphilips:
OK. yeah that is good enough reason to make it public I suppose.

PR #123, opened 2020-08-19
Fix chain hashes being serialized in reverse

canndrew:
This is a fix for changes introduced in https://github.com/joemphilips/DotNetLightning/pull/118 which caused chain hashes to get serialized in reverse order.

This allows DNL to open channels with lnd again.

### DIFF
--- a/tests/DotNetLightning.Core.Tests/Serialization.fs
+++ b/tests/DotNetLightning.Core.Tests/Serialization.fs
@@ -124,7 +124,7 @@ module SerializationTest =
 
                     let unsignedChannelAnnoucement = {
                         Features = features
-                        ChainHash = if (not nonbitcoinChainHash) then uint256(hex.DecodeData("6fe28c0ab6f1b372c1a6a246ae63f74f931e8365e15a089c68d6190000000000")) else uint256(hex.DecodeData("000000000933ea01ad0ee984209779baaec3ced90fa3f408719526f8d77f4943"))
+                        ChainHash = if (not nonbitcoinChainHash) then uint256(hex.DecodeData("000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f")) else uint256(hex.DecodeData("000000000933ea01ad0ee984209779baaec3ced90fa3f408719526f8d77f4943"))
                         ShortChannelId = ShortChannelId.FromUInt64(2316138423780173UL)
                         NodeId1 = NodeId(privKey1.PubKey)
                         NodeId2 = NodeId(privKey2.PubKey)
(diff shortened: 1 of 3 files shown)


knocte:
The commit message only has title.
Please sync it with the PR (which has title and body).

canndrew:
> The commit message only has title.

Oops. Fixed.

joemphilips:
Are you sure about this? In my understanding, LN protocol sends chain-hashes in big-endian. So read/writing it with `lendian=false` will result to reversed (i.e. little-endian, consensus-compatible)

Also, please add a test to show the behavior change.

knocte:
He's modifying the current tests, so his point is that they are wrong. (I'm not sure about this, but what I'm sure about is that your PR broke our end2end tests.)

tests/DotNetLightning.Core.Tests/Serialization.fs
```diff
@@ -124,7 +124,7 @@ module SerializationTest =
 
                     let unsignedChannelAnnoucement = {
                         Features = features
-                        ChainHash = if (not nonbitcoinChainHash) then uint256(hex.DecodeData("6fe28c0ab6f1b372c1a6a246ae63f74f931e8365e15a089c68d6190000000000")) else uint256(hex.DecodeData("000000000933ea01ad0ee984209779baaec3ced90fa3f408719526f8d77f4943"))
```
joemphilips, line 127: `uint256(hex.DecodeData("6fe28c0ab6f1b372c1a6a246ae63f74f931e8365e15a089c68d6190000000000"))`

Equals to `NBitcoin.Network.Main.Consensus.HashGenesisBlock` which means, you shouldn't change this line.
Instead change the expected value in the line 148.

joemphilips:
I've been digging into this.

It seems that I've been misunderstanding the encoding of the uint256 value.
I was assuming hash256 in bitcoin wire protocol is little-endian encoded, which was wrong. What is little-endian encoded is bitcoin-core internal representation and return value of the rpc call.
So I thought NBitcoin uses same internal representation with bitcoin-core, but indeed it is not. It uses big-endian encoding for BlockHash.

I suspect we should change all other occurrances of `ChainHash` field (e.g. `QueryChannelRangeMsg` , `InitTLV`) in the same way. Would you mind changing it? @canndrew

joemphilips:
ref: https://github.com/lightningnetwork/lightning-rfc/issues/237

tests/DotNetLightning.Core.Tests/Serialization.fs
```diff
@@ -143,7 +143,7 @@ module SerializationTest =
                     let mutable expected = hex.DecodeData("d977cb9b53d93a6ff64bb5f1e158b4094b66e798fb12911168a3ccdf80a83096340a6a95da0ae8d9f776528eecdbb747eb6b545495a4319ed5378e35b21e073a1735b6a427e80d5fe7cd90a2f4ee08dc9c27cda7c35a4172e5d85b12c49d4232537e98f9b1f3c5e6989a8b9644e90e8918127680dbd0d4043510840fc0f1e11a216c280b5395a2546e7e4b2663e04f811622f15a4f91e83aa2e92ba2a573c139142c54ae63072a1ec1ee7dc0c04bde5c847806172aa05c92c22ae8e308d1d2692b12cc195ce0a2d1bda6a88befa19fa07f51caa75ce83837f28965600b8aacab0855ffb0e741ec5f7c41421e9829a9d48611c8c831f71be5ea73e66594977ffd")
                     expected <- Array.append expected (hex.DecodeData("0000"))
                     if nonbitcoinChainHash then
-                        expected <- Array.append expected (hex.DecodeData("43497fd7f826957108f4a30fd9cec3aeba79972084e90ead01ea330900000000"))
+                        expected <- Array.append expected (hex.DecodeData("000000000933ea01ad0ee984209779baaec3ced90fa3f408719526f8d77f4943"))
```
joemphilips, line 146: nit: You can just `Array.rev`

joemphilips:
> So I thought NBitcoin uses same internal representation with bitcoin-core, but indeed it is not. It uses big-endian encoding for BlockHash.

Erk no. NBitcoin uses little-endian to hold `uint256`  I really have to clarify this.

joemphilips:
I've been trying to make things clear in https://github.com/lightningnetwork/lightning-rfc/issues/795

joemphilips:
So the conclusion is

* "little/big-endian" notation is confusing. we should call "non-reversed/reversed". non-reversed is the format in the wire protocol for both bitcoin layer1 and LN (e.g. "chain_hash"). reversed is the format which bitcoin-core rpc uses (e.g. "chain_id").
* So the change here is correct. NBitcoin refers "non-reversed" 256bit value as "little-endian",  so we must not flip when reading bytes.

> I suspect we should change all other occurrances of ChainHash field (e.g. QueryChannelRangeMsg , InitTLV) in the same way. Would you mind changing it?

So would you address this comment above? @canndrew then I will merge.

canndrew:
I've pushed a commit to reverse the byte order of chain hashes in TLVs. @joemphilips: you mentioned `QueryChannelRangeMsg` as something needing to be fixed. Can you elaborate on that?

joemphilips:
@canndrew Yes, all fields which contains `chain_hash` are named `ChainHash` , but there is one exception. `InitTLV` which holds chain hash information in `Network` case.

Speaking about `QueryChannelRangeMsg`,  it seems you have already fixed the `ChainHash` field, so nothing is left do.

joemphilips:
Oh you have already fixed `InitTLV`, sorry I didn't notice.

Please address this comment, otherwise LGTM.
https://github.com/joemphilips/DotNetLightning/pull/123/files#r473561145

tests/DotNetLightning.Core.Tests/Serialization.fs
canndrew: Is there a reason not to write it in that order though? This is the order that's expected on the wire, and the order that `uint256.Parse` accepts.

canndrew: For instance, line 148 (the other branch of this `if` expression) uses that order.

joemphilips: Not really, yeah I've changed my mind, my concern was kind of bikeshedding.
 I will just merge as is.

PR #125, opened 2020-08-24
Core: simplify types wrt serialization

knocte:
These members, being non-static, public and having no paramters, ended
up used in serialization. However there's no need for them to be included
because they can be recalculated later from the serialized form, plus
it reduces their marshalling size.

PS: Joe DO NOT MERGE this yet, I want to discuss something first.


src/DotNetLightning.Core/Channel/Commitments.fs
```diff
@@ -110,7 +110,7 @@ type RemoteNextCommitInfo =
             (fun commitmentPubKey remoteNextCommitInfo ->
                 match remoteNextCommitInfo with
                 | Waiting _ -> remoteNextCommitInfo
-                | Revoked commitmentPubKecommitmentPubKey -> Revoked commitmentPubKey)
+                | Revoked _commitmentPubKecommitmentPubKey -> Revoked commitmentPubKey)
```
knocte, line 113: @joemphilips note on this line change above ^ not only the variable name is weird (we should fix it)... is it also a bug? I mean maybe the result should be `Revoked commitmentPubKecommitmentPubKey` (or whatever new var name we choose) instead of `Revoked commitmentPubKey`; let me know what you think

src/DotNetLightning.Core/Serialize/Features.fs
```diff
@@ -296,6 +298,6 @@ type FeatureBit private (bitArray) =
     interface IComparable with
         member this.CompareTo(o) =
             match o with
-            | :? FeatureBit as fb -> this.CompareTo(fb)
+            | :? FeatureBits as fb -> this.CompareTo fb
             | _ -> -1
```
canndrew, line 302: Is it normal to return `-1` when two things can't be compared due to being incompatible types? Or should this be an exception?

knocte: this is actually how comparisons work in .NET: the CompareTo methods have to return either -1, 1, or 0 (the latter is if they are the same element).

src/DotNetLightning.Core/Channel/Commitments.fs
joemphilips: It is not a bug but you are right the variable name is weird. `commitmentPubKecommitmentPubKey` should be just `_` as in the case for in the `Waiting_`

knocte: cool thanks I'll rename

joemphilips: I don't know why I ended up with this nonsense variable name. Probably  the editor fooled me.

src/DotNetLightning.Core/Serialize/Features.fs
canndrew: I get that the function should normally return -1 / 0 / 1 to indicate less / equal / greater, I'm just wondering if a `FeatureBits` should be considered less than any random `obj` you pass to this function or if it's an error to compare them at all.

knocte: oh I don't know TBH

joemphilips: Hah, indeed, official examples throws `ArgumentException` .  Maybe it is good to fix.

https://docs.microsoft.com/en-us/dotnet/api/system.icomparable?redirectedfrom=MSDN&view=netcore-3.1

knocte: ok cool I'll add a fix for this while I'm at it

knocte:
@joemphilips this should be ready now

PR #126, opened 2020-09-03
Support Aezeed

joemphilips:
package `AEZ` is a direct port of https://github.com/Yawning/aez
Currently it does not support SIMD or other magics for speed.

AEZ is necessary building block for [aezeed](https://github.com/lightningnetwork/lnd/tree/master/aezeed) which intends to replace the BIP39.

My first plan was to put this to NBitcoin because strictly speeking aez is not a LN-specific thing.
But for following reasons I will probably place it to here.
1. NBitcoin's BouncyCastle does not include `Blake2b` which is a necessary building block for aez.
2. I did not want to deal with NBitcoin's compiler flags (e.g. `IF_SPAN`)


joemphilips:
AFAIK, lnd uses aez for 1. seed-backup with aezeed, 2. encypting channel state before saving it into disk. I'm planning to do the same in NRustLightning

DotNetLightning.sln
```diff
@@ -29,22 +29,22 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "NSec.Cryptography", "src\NS
...
-Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "DotNetLightning.Client", "src\DotNetLightning.Client\DotNetLightning.Client.fsproj", "{07862556-0436-4B07-A23B-496CAEFEDCE7}"
-EndProject
 Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "ResultUtils", "src\ResultUtils\ResultUtils.fsproj", "{21ECBAC8-A75F-4103-81F0-A8E135BE5AC6}"
 EndProject
 Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "TaskUtils", "src\TaskUtils\TaskUtils.fsproj", "{C21C0204-736B-4A66-A3E4-608002C91505}"
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "InternalBech32Encoder", "src\InternalBech32Encoder\InternalBech32Encoder.csproj", "{C7881917-9F0B-485B-9F33-8A58E1084B3F}"
 EndProject
-Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "DotNetLightning.Client.Tests", "tests\DotNetLightning.Client.Tests\DotNetLightning.Client.Tests.fsproj", "{3070F200-DDF7-4969-9967-9F33E1729361}"
-EndProject
```
knocte, line 41: these two hunks above look unrelated?

joemphilips: I've moved it into separate PR. #127

src/DotNetLightning.Core/Crypto/CryptoUtils.fs
```diff
@@ -285,7 +285,7 @@ module BouncyCastle =
 
     type CryptoImpl() =
         interface ICryptoImpl with
-            member this.newSecp256k1() = BouncySecp256k1() :> ISecp256k1
+            member this.newSecp256k1() = new BouncySecp256k1() :> ISecp256k1
```
knocte, line 288: @joemphilips note this is already fixed in my PR

src/DotNetLightning.Core/Crypto/RevocationSet.fs
```diff
@@ -47,7 +47,7 @@ type RevocationSet private (keys: list<CommitmentNumber * RevocationKey>) =
                             sanityCheck tail
         let commitmentNumbers, _ = List.unzip keys
         if not (sanityCheck commitmentNumbers) then
-            failwith "commitment number list is malformed: %A" commitmentNumbers
+            failwithf "commitment number list is malformed: %A" commitmentNumbers
```
knocte, line 50: this too

joemphilips:
@knocte Would you mind helping me with fixing mono build?  I now use "BouncyCastle" flag in two projects, and one of them ( `AEZ.csproj` ) fails in mono build. (It seems it is failng to find the BouncyCastle package)
I also tried to move some duplicated logics into top-level `Directory.Build.props`, I thought this might help for mono build but it seems it doesn't.

knocte:
Mmm I'm not sure what could be wrong, but why add PrivateAssets attrib to Portable.BouncyCastle PackageRef?

The reason I'm looking at that ref is because this build error seems to suggest that somehow it's not founding that dependency:
```
(CoreCompile target) -> 
  IBlake2B.cs(21,25): error CS0246: The type or namespace name 'Org' could not be found (are you missing a using directive or an assembly reference?) [/home/runner/work/DotNetLightning/DotNetLightning/src/AEZ/AEZ.csproj]
```

joemphilips:
@knocte 
It seems `-p:BouncyCastle` flag is needed during restoration. `nuget` command did not take any option so I used `dotnet nuget` instead.

I've also removed legacy `DotNetLightning.Server` and  its tests. I'm pretty sure no one uses this , but I can create separate pr if you like (it is only for dotnet core anyway)

joemphilips:
I will merge this in 1 or 2 days.

knocte:
@joemphilips last commit has a typo: serer -> server

@canndrew hey double check that we don't need DNL.Server for anything :) thanks!

Directory.Build.props
```diff
@@ -6,6 +6,26 @@
...
+      <PropertyGroup>
+        <OtherFlags>$(OtherFlags) -d:BouncyCastle</OtherFlags>
+        <DefineConstants>$(DefineConstants);BouncyCastle</DefineConstants>
+      </PropertyGroup>
+    </When>
+  </Choose>
+
+  <ItemGroup>
+    <PackageReference Include="NBitcoin" Version="5.0.53" />
+    <PackageReference Condition="'$(BouncyCastle)'=='true'" Include="Portable.BouncyCastle" Version="1.8.6.7" PrivateAssets="all" />
```
knocte, line 26: I'd still ask why use PrivteAssets for a nuget reference?

src/AEZ/AEZ.csproj
```diff
@@ -0,0 +1,22 @@
...
+        </AssemblyAttribute>
+    </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Condition="'$(BouncyCastle)'!='true'" Include="..\NSec\Cryptography\NSec.Cryptography.csproj" PrivateAssets="all" />
+  </ItemGroup>
+
+    <ItemGroup>
+      <PackageReference Include="System.Memory" Version="4.5.3" />
+      <PackageReference Condition="'$(BouncyCastle)'=='true'" Include="Portable.BouncyCastle" Version="1.8.6.7" PrivateAssets="all" />
```
knocte, line 20: same here

Directory.Build.props
joemphilips: I once wanted to use internal projects (e.g. Macaroons, InternalBech32Encoders) from the consumer application. I required me to add `PrivateAssets="all"`

But TBH I don't understand why this solved my problem. I should do some research.
And be sure if really need it for `BouncyCastle`

knocte: Macaroons, InternalBech32Encoders, are projects whose sources are here in DNL repo, they are not nuget references. AFAIK we add PrivateAssets attrib to them so that their DLLs are bundled in our nuget package. But BouncyCastle is a nuget dependency and doesn't need to be bundled, i.e. it should just be listed as a dep in nuget.org website.

joemphilips: Yeah you are right. I've just fixed.

knocte:
build is broken on macOS

joemphilips:
> build is broken on macOS

Yeah, I've been looking into it.
This is super weird. It does not reproduce on my local macos. And it is not even a property test.

joemphilips:
I'm going to merge this.
Tell me if I have to revert for the deletion of the `DotNetLighnitng.Server` (I'm quite certain that there is no need, though.)

PR #127, opened 2020-09-04
remove useless Client and its tests

joemphilips:
These were the remainder of my first attempt.
In which I tried to build the full LN node from scrtach.
But since it was blocking the CI, there are no reason to
maintain anymore.


knocte:
@canndrew can you review to make sure we don't need any of this?

canndrew:
@knocte: :+1:  We don't use the Client stuff at all.

knocte:
@joemphilips you forgot to update the Dependency Graph image from the README though ;)

joemphilips:
> you forgot to update the Dependency Graph image from the README though ;)

Yeah, I also removed `DotNetLighting.Server`  too. I will update readme in another PR.
Since I'm now 100% sure that my [NRustLightning](https://github.com/joemphilips/NRustLightning) works, I will probably rebrand this repository as a utility library for working with lightning-related operation.

PR #129, opened 2020-09-11
Refactor Macaroons and its tests

joemphilips:
Previously, user had to call `initMacaroon` before using 3rd party
Caveats, This was a temporary wrokaround before preparing compiler
flag at solution level, but now we have it. We can switch which build
it uses on `Macaroons/` side.
So move initialization logic to CryptoAlgorithm and switch on macaroons
tests in Macaroon side.

We still do not have 3rd-party caveats support for BouncyCastle build.
This is because, BouncyCastle does not support XSalsa20-Poly1305
(a.k.a. `crypto_secret_box` in NaCl).


knocte:
> We still do not have 3rd-party caveats support for BouncyCastle build.

What do you mean with this sentence?

DotNetLightning.sln.DotSettings
```diff
@@ -7,6 +7,7 @@
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Bitcoind/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Bitcoin_0027s/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=blockchain/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=bouncycastle/@EntryIndexedValue">True</s:Boolean>
```
knocte, line 10: should we delete this file from the repo and add it to gitignore? I think it's some IDE thing

src/Macaroons/Macaroon.cs
```diff
@@ -12,7 +12,12 @@ namespace Macaroons
     public class Macaroon
     {
         
-        public static CryptoAlgorithm Crypto = new DummyCryptoAlgorithm();
+        /// <summary>
+        /// Cryptographic algorithm used for 3rd caveats Encryption/Decryption.
```
knocte, line 16: missing word "party"?

joemphilips:
> What do you mean with this sentence?

Macaroon has a feature called third-party-cavetas, in which macaroon holder can restrict its  access rights before delegating it to someone else.
It requires cryptographically secure encryption/decryption algorithm to work fine.
In the reference implementation in C, it uses xsalsa20-poly1305. So test vector is defined for that.

I thought about using ChaCha20-Poly1305 for BouncyCastle build. But I'm not sure I should do this. because if the user switches BouncyCastle build to non-BouncyCastle build at somepoint, the previous macaroon will be invalidated without making user aware of it.

(Note that 3rd-party caveats are advanced usage and LSAT can work fine without it)

DotNetLightning.sln.DotSettings
joemphilips: I sometimes work in different machine so this helps me.
It does not hurt and might help someone who want to work with IDE, so I want to keep it here.

joemphilips:
After thinking about this a bit, I think I should just use ChaCha20-Poly1305 for both build.
It requires me to keep `SecretBoxCryptoAlgorithm` for passing some test vectors tough.

joemphilips:
Finished.
I'm going to merge this in few days.

Issue #131, opened 2020-10-16
Channel closing fails with "failed to finalize psbt"


canndrew:
## Steps to reproduce:

0. (In geewallet we have a test that does this):
1. opens a channel against another geewallet
2. closes the channel (using DotNetLightning)

## Current results:

About 50% of the time the peer which accepts the channel close (not the initiator) will crash with an exception thrown from DotNetLightning:

```
TransactionRelatedErrors
        [FailedToFinalizeScript
           "failed to finalize psbt Errors: seq [Input 0: The finalized input script does not properly validate "NullFail"] 
 PSBTInput: {
  "index": 0,
  "partial_signatures": {
    "02db5e2701d93ca1b07c851dda2125d4b6cc36fe378db3719f384ba7dd2dfa2d8c": "304402206b74ee87e5955097515c9c300b01a9b4f7762ac102915291add469cb3dcf256c022073457a0a6f931a0ff29e9fbf29972d3a2d9b2147946b09839cf5a3ecd018dfb701",
    "0389ecd0689e9db416e2a04eda4cc817aae00bb906765f29616617ec10120039cd": "304402205e3c21be6cd331a590662d8161c521e0e82440f0981959cc7f5a43b72e2c2c5d022040de0b8d750178fbc81b339368a287142b7e0789971f0df0b1e659c71f74a22101"
  },
  "witness_script": "2 02db5e2701d93ca1b07c851dda2125d4b6cc36fe378db3719f384ba7dd2dfa2d8c 0389ecd0689e9db416e2a04eda4cc817aae00bb906765f29616617ec10120039cd 2 OP_CHECKMULTISIG",
  "witness_utxo": {
    "value": "0.10000000",
    "scriptPubKey": "0 97548049d3dea0202df8566d01fcd747b1157b9c4bd5b49aa07e1f49a4883ba5"
  },
  "bip32_derivs": []
} 
[...]

PR #132, opened 2020-10-20
Fix closing tx output-reordering bug

canndrew:
The lightning spec decrees that the closing tx's outputs need to be in lexicographical order. Instead, they were ending up in the wrong order thanks to (a) not being added in the correct order and (b) `TransactionBuilder` shuffling them anyway when calling `BuildTransaction`.

`makeClosingTx` now makes the closing tx using `Transaction` directly rather than `TransactionBuilder`. This allows us to set the order of the outputs explicitly.

### DIFF
--- a/src/DotNetLightning.Core/Transactions/Transactions.fs
+++ b/src/DotNetLightning.Core/Transactions/Transactions.fs
@@ -1,6 +1,7 @@
 namespace DotNetLightning.Transactions
 
 open System
+open System.IO
 open System.Linq
 
 open NBitcoin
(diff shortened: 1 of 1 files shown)


knocte:
@canndrew good work! can you add the string "Fixes https://github.com/joemphilips/DotNetLightning/issues/131" in the commit message please? This way we have a reference to the filed bug (and github closes it automatically when the PR is merged).

src/DotNetLightning.Core/Transactions/Transactions.fs
```diff
@@ -735,19 +736,29 @@ module Transactions =
...
+                    txIn.Sequence <- !> 0xffffffffu
+                    txIn.PrevOut <- commitTxInput.Outpoint
+                    txIn
+                let tx = Transaction.Create n
                 tx.Version <- 2u
-                tx.Inputs.[0].Sequence <- !> 0xffffffffu
+                tx.LockTime <- !> 0u
+                tx.Inputs.Add txIn
+                for txOut in outputs do
+                    tx.Outputs.Add(txOut)
```
knocte, line 761: nit: no need for parenthesis here

src/DotNetLightning.Core/Transactions/Transactions.fs
```diff
@@ -735,19 +736,29 @@ module Transactions =
...
-                           .AddCoins(commitTxInput)
-                           .SendFees(closingFee)
-                           .SetLockTime(!> 0u)
-                for (money, dest) in outputs do
-                    txb.Send(dest, money) |> ignore
-                let tx =  txb.BuildTransaction(false)
+                let txIn =
+                    let txIn = TxIn()
+                    txIn.Sequence <- !> 0xffffffffu
+                    txIn.PrevOut <- commitTxInput.Outpoint
```
knocte, line 754: rather initialize it without mutation `let txIn = TxIn(PrevOut = commitTxInput.Outpoint, Sequence = !> 0xffffffffu)`

knocte: also put 0xffffffffu under a constant (so that we know what it refers to)

src/DotNetLightning.Core/Transactions/Transactions.fs
```diff
@@ -735,19 +736,29 @@ module Transactions =
                 else
                     spec.ToLocal.ToMoney(), spec.ToRemote.ToMoney() - closingFee
 
-            let maybeToLocalOutput = if (toLocalAmount >= dustLimit) then Some(toLocalAmount, localDestination) else None
-            let maybeToRemoteOutput = if (toRemoteAmount >= dustLimit) then Some(toRemoteAmount, remoteDestination) else None
-            let outputs = seq [maybeToLocalOutput; maybeToRemoteOutput] |> Seq.choose id
+            let maybeToLocalOutput = if (toLocalAmount >= dustLimit) then Some(TxOut(toLocalAmount, localDestination)) else None
+            let maybeToRemoteOutput = if (toRemoteAmount >= dustLimit) then Some(TxOut(toRemoteAmount, remoteDestination)) else None
```
knocte, line 740: now that you're changing these 2 lines above, IMO it's better to split them, to not have so much logic in one line; like this (and removing unneeded parenthesis):

```
let maybeToLocalOutput =
    if toLocalAmount >= dustLimit then
        Some(TxOut(toLocalAmount, localDestination))
    else
        None
```

canndrew:
Closing in favor of https://github.com/joemphilips/DotNetLightning/pull/133

PR #133, opened 2020-10-20
Fix closing tx output-reordering bug

knocte:
(Based on PR[1] from Andrew Cann.)

The lightning spec decrees that the closing tx's outputs need to be in
lexicographical order. Instead, they were ending up in the wrong order
thanks to (a) not being added in the correct order and (b)
`TransactionBuilder` shuffling them anyway when calling
`BuildTransaction`.

`makeClosingTx` now uses an explicit order and we disable shuffling
inputs and outputs for all transactionBuilders (this feature was
introduced in NBitcoin 5.0.54 [2] so an upgrade to this version is
included).

(As an extra improvement, this commit also renames 'n' to 'network',
and creates a constant for 0xffffffffu.)

Fixes https://github.com/joemphilips/DotNetLightning/issues/131

[1] https://github.com/joemphilips/DotNetLightning/pull/132
[2] https://github.com/MetacoSA/NBitcoin/commit/db0d8247d372c2aa5b23017c038618ddbf765241

(Supersedes https://github.com/joemphilips/DotNetLightning/pull/132 )

### DIFF
--- a/src/DotNetLightning.Core/Transactions/Transactions.fs
+++ b/src/DotNetLightning.Core/Transactions/Transactions.fs
@@ -1,6 +1,7 @@
 namespace DotNetLightning.Transactions
 
 open System
+open System.IO
 open System.Linq
 
 open NBitcoin
@@ -297,6 +298,13 @@ module Transactions =
         [<Literal>]
         let OFFERED_HTLC_SCRIPT_WEIGHT = 133uy
 
+    let private createTransactionBuilder (network: Network) =
+        let txb = network.CreateTransactionBuilder()
+        txb.ShuffleOutputs <- false
+        txb.ShuffleInputs <- false
+        txb
+
+    let UINT32_MAX = 0xffffffffu
 
     let private trimOfferedHTLCs (dustLimit: Money) (spec: CommitmentSpec): DirectedHTLC list =
         let htlcTimeoutFee = spec.FeeRatePerKw.CalculateFeeFromWeight(HTLC_TIMEOUT_WEIGHT)
@@ -356,7 +364,8 @@ module Transactions =
                      (localHTLCPubKey: PubKey)
                      (remoteHTLCPubkey: PubKey)
                      (spec: CommitmentSpec)
-                     (n: Network)=
+                     (network: Network)
+                         =
         let commitFee = commitTxFee localDustLimit spec
         let (toLocalAmount, toRemoteAmount) =
             if (localIsFunder) then
@@ -392,7 +401,7 @@ module Transactions =
         let lockTime = obscuredCommitmentNumber.LockTime
 
         let tx =
-            let tx = n.CreateTransaction()
+            let tx = network.CreateTransaction()
             tx.Version <- 2u
             let txin = TxIn(inputInfo.Outpoint)
             txin.Sequence <- sequence
@@ -405,7 +414,7 @@ module Transactions =
             tx.LockTime <- lockTime
             tx
         let psbt =
-            let p = PSBT.FromTransaction(tx, n)
+            let p = PSBT.FromTransaction(tx, network)
             p.AddCoins(inputInfo)
         { CommitTx.Value = psbt; WhichInput = 0 }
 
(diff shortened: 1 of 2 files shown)


src/DotNetLightning.Core/Transactions/Transactions.fs
```diff
@@ -297,6 +298,13 @@ module Transactions =
         [<Literal>]
         let OFFERED_HTLC_SCRIPT_WEIGHT = 133uy
 
+    let CreateTransactionBuilder (network: Network) =
```
joemphilips, line 301: `let` function must be camelCase. https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting#use-camelcase-for-module-bound-public-functions

Or in this case, you can define as an extension method to the `Network` , in that case you should define in `Utis/NBitcoinExtensions.fs`

src/DotNetLightning.Core/Transactions/Transactions.fs
```diff
@@ -297,6 +298,13 @@ module Transactions =
...
+        let txb = network.CreateTransactionBuilder()
+        txb.ShuffleOutputs <- false
+        txb.ShuffleInputs <- false
+        txb
+
+    let UINT32_MAX = 0xffffffffu
```
joemphilips, line 307: 1. the name is confusing, because `UInt32.MaxValue` is `0xffffffff`, what about `UINT32_MAX_MINUS_1`
2. Using `[<Literal>]` will improve the performance a bit.

src/DotNetLightning.Core/Transactions/Transactions.fs
```diff
@@ -735,20 +772,31 @@ module Transactions =
...
+                        yield (toLocalAmount, localDestination)
+                    if toRemoteAmount >= dustLimit then
+                        yield (toRemoteAmount, remoteDestination)
+                }
+                |> Seq.sortBy (fun (money, dest) ->
+                    let txOut = TxOut(money, dest)
+                    use memoryStream = new MemoryStream()
+                    let stream = NBitcoin.BitcoinStream(memoryStream, true)
+                    txOut.ReadWrite stream
+                    memoryStream.ToArray()
```
joemphilips, line 787: Isn't this same with `txOut.ToBytes()` ?

src/DotNetLightning.Core/Transactions/Transactions.fs
joemphilips: Oh never mind I misread.

src/DotNetLightning.Core/Transactions/Transactions.fs
knocte: why do you want to differ from the Microsoft's .NET API design guidelines? In them, methods are always PascalCase, not camelCase.

src/DotNetLightning.Core/Transactions/Transactions.fs
knocte: I don't know, I copied this bit from @canndrew's PR. If it is, you prefer ToBytes()?

src/DotNetLightning.Core/Transactions/Transactions.fs
joemphilips: > I can or I should? Cause I'm not a fan of extension methods, so if the former, I'll leave it as is.

I meant "can", No problem.

src/DotNetLightning.Core/Transactions/Transactions.fs
joemphilips: Yes, `ReadWrite` is very bitcoin-specific and the reader have to know about the context to understand. So I prefer `ToBytes`

src/DotNetLightning.Core/Transactions/Transactions.fs
knocte: >  (so I guess adding private or internal is better BTW)

Sure sorry I make it private.

> If we want to expose function to C#, then we can use

Why go into such annoying standard and not just adopt the same standards as C#? I only use camelCase myself for functions when they are local functions (functions inside of functions) or when the function is passed as a parameter. Personal preference though...

joemphilips: > Why go into such annoying standard and not just adopt the same standards as C#? I only use camelCase myself for functions when they are local functions (functions inside of functions) or when the function is passed as a parameter. Personal preference though...

I think it is kind of bikeshedding to chose which style, but consistency matters. So please use camelCase in this code base.

knocte:
Updated.

Issue #98, opened 2020-05-21
Stop supporting Infrastructure.


joemphilips:
I think I will stop running infrastructure tests in CI. It is too dirty that I might re-write everything from scratch while debugging its flaky behavior.

Recently, I’ve been spending most of my efforts on [NRustLightning](https://github.com/joemphilips/NRustLightning), and if this works fine (it probably does.) I don’t need DotNetLightning.Infastructure anymore.
And @knocte and @canndrew are not planning to use it either. So there is no reason to maintain it anymore.

knocte:
> Recently, I’ve been spending most of my efforts on NRustLightning, and if this works fine (it probably does.) I don’t need DotNetLightning.Infastructure anymore.
And @knocte and @canndrew are not planning to use it either. So there is no reason to maintain it anymore.

By the same reasoning above, do you agree to nuke the non-BouncyCastle build?

joemphilips:
By the same reasoning above, do you agree to nuke the non-BouncyCastle build?

Yes.

canndrew:
This sounds fine to me. We're not using the infrastructure stuff at all.

PR #134, opened 2020-10-21
Remove Infrastructure(.*)

knocte:
Fixes https://github.com/joemphilips/DotNetLightning/issues/98


joemphilips:
Yeah, I did not remove infrastructure because it may be good example for how to use Channels and Peers, but I guess making code base small has more benefits.

knocte:
> Yeah, I did not remove infrastructure because it may be good example for how to use Channels and Peers, but I guess making code base small has more benefits.

Well it's still there in git history, maybe extract some code snippets from it into the Readme?

joemphilips:
> maybe extract some code snippets from it into the Readme?

Nah, it will quickly become unmaintainable, ideally we should have tests for Channels and Peers instead.

PR #135, opened 2020-10-21
[RFC/WIP] Core: make sure RBF is enabled

knocte:
Lightning transactions should always be RBF enabled, because the
miner fee rates can vary wildly since the moment they are signed
until the moment they are broadcasted.

### DIFF
--- a/src/DotNetLightning.Core/Transactions/Transactions.fs
+++ b/src/DotNetLightning.Core/Transactions/Transactions.fs
@@ -302,6 +302,7 @@ module Transactions =
         let txb = network.CreateTransactionBuilder()
         txb.ShuffleOutputs <- false
         txb.ShuffleInputs <- false
+        txb.OptInRBF <- true
         txb
 
     let UINT32_MAX = 0xffffffffu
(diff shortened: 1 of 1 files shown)


joemphilips:
All LN involved TX specifies `sequence` value explicitly in BOLT03, so we don't need this. In fact in the case of TX which requires two signatures, we only exchange the signature and not TX itself so the shape of the TX must be precisely defined.

joemphilips:
Miner fee problem is addressed recently by `option_anchor_output` feature, which is not reflected in this library yet :(

joemphilips:
See [bitcoin-optech newsletter](https://bitcoinops.org/en/topics/anchor-outputs/#:~:text=Anchor%20outputs%20are%20special%20outputs,signed%20by%20the%20participating%20parties.) for `option_anchor_outputs`

knocte:
> Miner fee problem is addressed recently by option_anchor_output feature

Anchor outputs currrently has a limitation/bug: https://lists.linuxfoundation.org/pipermail/lightning-dev/2020-September/002796.html TL/DR: watchtowers are ineffective.

Issue #99, opened 2020-05-22
Mutable collections are causing race conditions


knocte:
See https://github.com/joemphilips/DotNetLightning/runs/695486967

(We should use Map instead of ConcurrentDictionary.)

joemphilips:
I'm not 100% sure if the mutable collection is what causing this.
It may be something specific in Infrastructure tests. Anyway, #98 might be a solution if it is too hard to debug.

knocte:
I'd like to know Andrew's opinion on getting rid of Infrastructure project and tests before we nuke it. However, regardless of that, I think switching to immutable collections is still a worthy change.

PR #136, opened 2020-10-27
Key types refactor

### LINKED ISSUE #130: DotNetLightning conflates RevocationKey and PerCommitmentSecret
When trying to use DotNetLightning to implement revocation in geewallet I struggled to make sense of the revocation APIs until I realized they don't actually make sense. The main problem is that revocation keys and per commitment secrets are two different things, yet the `RevocationKey` type is actually a type that holds a per commitment secret. This is my doing, since I added the `RevocationKey`/`RevocationSet` types. Additionally, some key-related fields and variables have names that don't match the names in the lightning spec. This made it difficult for me to be confident that I'm using keys correctly, especially after I became aware that at least one thing I had to use was completely misnamed as another thing.

To fix this I'm working on a PR that (a) renames `Revocation{Key,Set}` as well as some fields to match the names used in the spec and (b) adds `Key`/`PubKey`-wrapper types for some more keys (like I did for `RevocationKey`) so that it's a type-error to use them incorrectly (eg. by using the wrong key to derive some other key). This should make both DNL's and its consumers' code more readable as well as more bug-proof.

The WIP branch is here: https://github.com/canndrew/DotNetLightning/tree/key-types-refactor

canndrew:
Fixes #130

This PR adds a bunch of new keys types, renames some of the existing types and refactors some of the key-handling code.

In summary:

* `CommitmentPubKey` and `RevocationKey` have been renamed to `PerCommitmentPoint` and `PerCommitmentSecret`. This is the terminology used in the lightning spec. Also it's not correct to call the per commitment secret a "revocation key" because a revocation key is something entirely different.

* Add public/private key types and public/private basepoint types for payment keys, delayed payment keys, revocation keys, and htlc keys. Also make funding public/private keys their own type.

* Add all of the above under a new `Crypto.Keys` module, and move a few other key-related things into that module since they're required by the methods in that module.

* Use the new key types whereever possible, rather than using more general key types. eg. transaction creation and signing functions now take keys of the required specific types.

* Remove `IKeysRepository` and just use `ChannelPrivKeys` (formerly `ChannelKeys`) instead. `IKeysRepository` only holds the keys for a single channel, which is also what `ChannelPrivKeys` does. So having both was redundant.

The messages of the individual commits go into more detail of the changes behind each commit.


src/DotNetLightning.Core/Crypto/Keys.fs
```diff
@@ -0,0 +1,584 @@
...
+        static member Add(lhs: Key, rhs: Key): Key =
+            let lhsBytes = lhs.ToBytes()
+            let rhsBytes = rhs.ToBytes()
+            let tweak = ReadOnlySpan(lhsBytes)
+            match Secp256k1.PrivateKeyTweakAdd(tweak, rhsBytes.AsSpan()) with
+            | true -> Key rhsBytes
+            | false -> failwithf "failed to add Keys"
+
+    type PubKey with
+        member this.ExpandToBytes(): array<byte> =
```
joemphilips, line 40: I think this should be private.

src/DotNetLightning.Core/Crypto/Keys.fs
```diff
@@ -0,0 +1,584 @@
...
+            let compressedPubKeyBytes = this.ToBytes()
+            let compressedPubKeyBytesSpan = ReadOnlySpan(compressedPubKeyBytes)
+            match Secp256k1.PublicKeyParse(compressedPubKeyBytesSpan) with
+            | true, expandedPubKeyBytes -> expandedPubKeyBytes
+            | false, _ -> failwithf "Failed  to expand public key %A" this
+
+        static member CompressFromBytes(expandedPubKeyBytes: array<byte>): PubKey =
```
joemphilips, line 47: This too.

joemphilips (changes requested):

knocte:
This fixes issue 130? Please add the "Fixes ..." magic

src/DotNetLightning.Core/Channel/Channel.fs
```diff
@@ -633,25 +639,26 @@ module Channel =
...
                     cm.FundingScriptCoin.TxOut.Value - (cm.LocalCommit.PublishableTxs.CommitTx.Value.TotalOut)
                 do! checkRemoteProposedHigherFeeThanBefore lastCommitFeeSatoshi msg.FeeSatoshis
                 let! closingTx, closingSignedMsg =
-                    Closing.makeClosingTx (cs.KeysRepository, cm, state.LocalShutdown.ScriptPubKey, state.RemoteShutdown.ScriptPubKey, msg.FeeSatoshis, cm.LocalParams.ChannelPubKeys.FundingPubKey, cs.Network)
+                    Closing.makeClosingTx (cs.ChannelPrivKeys, cm, state.LocalShutdown.ScriptPubKey, state.RemoteShutdown.ScriptPubKey, msg.FeeSatoshis, cm.LocalParams.ChannelPubKeys.FundingPubKey, cs.Network)
                     |> expectTransactionError
-                let! finalizedTx = Transactions.checkTxFinalized closingTx.Value (closingTx.WhichInput) (seq [ cm.RemoteParams.FundingPubKey, TransactionSignature(msg.Signature.Value, SigHash.All) ]) |> expectTransactionError
+                let! finalizedTx = Transactions.checkTxFinalized closingTx.Value (closingTx.WhichInput) (seq [ remoteChannelKeys.FundingPubKey.RawPubKey(), TransactionSignature(msg.Signature.Value, SigHash.All) ]) |> expectTransactionError
```
knocte, line 661: nit: now that you're changing this line, let's split it because it's too long

tests/DotNetLightning.Core.Tests/TransactionBolt3TestVectorTests.fs
```diff
@@ -26,7 +26,7 @@ let log =
 let dataPath1 = Path.Join(AppDomain.CurrentDomain.BaseDirectory, "../../..", ("Data/bolt3-tx.json"))
 let data1 = dataPath1 |> File.ReadAllText |> JsonDocument.Parse
 
-let localPerCommitmentPoint = PubKey("025f7117a78150fe2ef97db7cfc83bd57b2e2c0d0dd25eaf467a4a1c2a45ce1486")
+let localPerCommitmentPoint = PerCommitmentPoint <| PubKey("025f7117a78150fe2ef97db7cfc83bd57b2e2c0d0dd25eaf467a4a1c2a45ce1486")
```
knocte, line 27: the commit message of this commit looks like it's fixing something, but you're actually changing the tests; how can this be? does the channel pass before and after this commit?

src/DotNetLightning.Core/Crypto/Keys.fs
```diff
@@ -53,8 +79,7 @@ and [<Struct>] PerCommitmentPoint(pubKey: PubKey) =
     member this.ToBytes(): array<byte> =
         this.RawPubKey().ToBytes()
 
-[<Struct>]
-type CommitmentSeed(lastPerCommitmentSecret: PerCommitmentSecret) =
+and [<Struct>] CommitmentSeed(lastPerCommitmentSecret: PerCommitmentSecret) =
```
knocte, line 563: you don't need to declare CommitmentSeed with `and`, you can just move it earlier than `ChannelKeys`

src/DotNetLightning.Core/Serialization/LightningStream.fs
```diff
@@ -325,20 +325,22 @@ type LightningReaderStream(inner: Stream) =
         let len = this.ReadUInt16(false)
         this.ReadBytes(int32 len)
 
+    member this.ReadKey() =
+        let b: array<byte> = this.ReadBytes(Key.BytesLength)
```
knocte, line 329: nit: no need for parens here, and now that you're gonna change this line, rename `b` to have a better name please

src/DotNetLightning.Core/Serialization/LightningStream.fs
```diff
@@ -325,20 +325,22 @@ type LightningReaderStream(inner: Stream) =
...
+        Key b
+
     member this.ReadPubKey() =
-        let b = this.ReadBytes(33)
+        let b = this.ReadBytes(PubKey.BytesLength)
```
knocte, line 333: no need for parenthesis here

tests/DotNetLightning.Core.Tests/KeyRepositoryTests.fs
```diff
@@ -96,14 +96,14 @@ let tests =
             let commitTx =
                 Transactions.makeCommitTx fundingScriptCoin
                                           CommitmentNumber.FirstCommitment
-                                          (localPubKeys.PaymentBasePubKey)
-                                          (remotePubKeys.PaymentBasePubKey)
+                                          (localPubKeys.PaymentBasepoint)
+                                          (remotePubKeys.PaymentBasepoint)
```
knocte, line 100: no need for parenthesis here

src/DotNetLightning.Core/Chain/KeysInterface.fs
```diff
@@ -60,7 +60,7 @@ type IKeysRepository =
     /// Must add signature to the PSBT *And* return the signature
     abstract member GetSignatureFor: psbt: PSBT * pubKey: PubKey -> TransactionSignature * PSBT
     /// Must add signature to the PSBT *And* return the signature
-    abstract member GenerateKeyFromBasepointAndSign: psbt: PSBT * pubkey: PubKey * basepoint: PubKey -> TransactionSignature * PSBT
+    abstract member GenerateKeyFromBasepointAndSign: psbt: PSBT -> pubkey: PubKey -> perCommitmentPoint: PerCommitmentPoint -> TransactionSignature * PSBT
```
knocte: "This makes the code simpler and easier to understand.", so this commit is just a refactoring? if so, it's better to use this word in the commit title

src/DotNetLightning.Core/Crypto/Keys.fs
```diff
@@ -248,6 +248,13 @@ and ChannelPubKeys = {
     HtlcBasepoint: HtlcBasepoint
 }
 
+and CommitmentPubKeys = {
```
knocte, line 324: no need for `and`, it's a new type and you're not using it from any type ahead (upper than) of it

src/DotNetLightning.Core/Chain/KeysInterface.fs
```diff
@@ -57,8 +57,8 @@ type IKeysRepository =
     /// restarted with some stale data.
     abstract member GetChannelKeys: inbound:bool -> ChannelKeys
 
-    /// Must add signature to the PSBT *And* return the signature
-    abstract member GetSignatureFor: psbt: PSBT * pubKey: PubKey -> TransactionSignature * PSBT
+    /// Must add funding pub key signature for to the PSBT *And* return the signature
+    abstract member SignWithFundingPrivKey: psbt: PSBT -> TransactionSignature * PSBT
```
knocte: "Making that explicit..." -> Sounds like this commit is a refactoring too (no change of behaviour), let's state it in the commit title

src/DotNetLightning.Core/Chain/KeysInterface.fs
```diff
@@ -60,7 +60,7 @@ type IKeysRepository =
     /// Must add funding pub key signature for to the PSBT *And* return the signature
     abstract member SignWithFundingPrivKey: psbt: PSBT -> TransactionSignature * PSBT
     /// Must add signature to the PSBT *And* return the signature
-    abstract member GenerateKeyFromBasepointAndSign: psbt: PSBT -> pubkey: PubKey -> perCommitmentPoint: PerCommitmentPoint -> TransactionSignature * PSBT
+    abstract member SignHtlcTx: psbt: PSBT -> perCommitmentPoint: PerCommitmentPoint -> TransactionSignature * PSBT
```
knocte: same here about the commit message, add "Refactoring: "

src/DotNetLightning.Core/Crypto/Keys.fs
```diff
@@ -239,6 +239,27 @@ type ChannelPrivKeys = {
...
+        let htlcPrivKey = perCommitmentPoint.DeriveHtlcPrivKey this.HtlcBasepointSecret
+        let htlcPubKey = htlcPrivKey.HtlcPubKey()
+        psbt.SignWithKeys(htlcPrivKey.RawKey()) |> ignore
+        match psbt.GetMatchingSig(htlcPubKey.RawPubKey()) with
+        | Some signature -> (signature, psbt)
+        | None ->
+            failwithf
+                "failed to get htlc signature for %A. with htlc pubkey (%A) and perCommitmentPoint (%A)"
+                psbt htlcPubKey perCommitmentPoint
+
```
knocte, line 314: it seems this hunk was not meant to go in this commit

src/DotNetLightning.Core/Channel/CommitmentsModule.fs
```diff
@@ -381,10 +380,10 @@ module internal Commitments =
                         | _ -> None
                     )
                 let localPerCommitmentSecret =
-                    chanPrivateKeys.CommitmentSeed.DerivePerCommitmentSecret cm.LocalCommit.Index
+                    channelPrivKeys.CommitmentSeed.DerivePerCommitmentSecret cm.LocalCommit.Index
                 let localNextPerCommitmentPoint =
                     let perCommitmentSecret =
-                        chanPrivateKeys.CommitmentSeed.DerivePerCommitmentSecret
+                        channelPrivKeys.CommitmentSeed.DerivePerCommitmentSecret
```
knocte, line 386: add "Refactoring: " to commit msg

src/DotNetLightning.Core/Crypto/Keys.fs
```diff
@@ -0,0 +1,584 @@
...
+            | false -> failwith "failed to multiplying PubKey by Key"
+
+        static member Add(lhs: PubKey, rhs: PubKey): PubKey =
+            let lhsBytes = lhs.ToBytes()
+            let rhsBytes = rhs.ToBytes()
+            match Secp256k1.PublicKeyCombine(lhsBytes.AsSpan(), rhsBytes.AsSpan()) with
+            | true, result -> PubKey.CompressFromBytes result
+            | false, _ -> failwith "failed to add PubKeys"
+
+type [<Struct>] FundingPubKey(pubKey: PubKey) =
```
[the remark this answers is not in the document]
canndrew: One reason is so that equality works. As reference types, two identical keys won't compare as equal if they're different references.

Also saving one layer of indirection seems like a good thing.

src/DotNetLightning.Core/Chain/KeysInterface.fs
```diff
@@ -47,146 +47,3 @@ type SpendableOutputDescriptor =
...
-                FundingPubKey = this.FundingKey.PubKey
-                RevocationBasePubKey = this.RevocationBaseKey.PubKey
-                PaymentBasePubKey = this.PaymentBaseKey.PubKey
-                DelayedPaymentBasePubKey = this.DelayedPaymentBaseKey.PubKey
-                HTLCBasePubKey = this.HTLCBaseKey.PubKey
-            }
-
-
-/// Interface to describe an object which can get user secrets and key material.
-type IKeysRepository =
```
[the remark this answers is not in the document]
canndrew: I don't understand what you mean. As it was, `ChannelPrivKeys` and `IKeysRepository` were doing pretty much the exact same thing. So I can't see the point in having both.

tests/DotNetLightning.Core.Tests/TransactionBolt3TestVectorTests.fs
canndrew: It does. But using a `PerCommitmentPoint` to represent a per commitment point is more correct. And it will be necessary for later commits which call methods on the `PerCommitmentPoint` directly rather than just converting it back to a `PubKey` to use it.

src/DotNetLightning.Core/Chain/KeysInterface.fs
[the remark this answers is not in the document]
canndrew: I think I get what you're saying now. Correct me if I'm wrong...

Holding private keys in RAM is dangerous because another process could read our memory and get the keys. `ChannelPrivKeys` is no different from `DefaultKeyRepository` in this regard (since they're pretty much the same type) however the keys repository was designed as an interface so that a user could replace it with something more secure.

I think if we want to take that design seriously then there's some things to consider:

Firstly, `IKeysRepository` shouldn't have an option to get the channel private keys at all, otherwise we'll end up holding them in RAM anyway. Even if the code only grabs the keys temporarily when it needs them we don't have a way to control when those keys get garbage collected. Also they'll still live on in RAM for a while after they get collected, assuming the CLR doesn't zero-out memory once it gets collected since that would be an expensive waste of time when the memory is being used for anything other than holding secrets.

Also, the user can't define an `IKeysRepository` that requires user intervention whenever DotNetLightning needs to access the keys. This is particularly true for the `NodeSecret` since we need it every time we so much as reply to a ping. But it's also true of the channel keys since you need them to create commitment transactions, unless (maybe?) if other nodes are willing to deal with a node which can't forward htlcs and which is extremely unresponsive when accepting payments.

We only actually need to hold the `NodeMasterKey`, rather than a hash map of private keys, since all the other private keys are derived from it.

So maybe a good design here would be to (a) reinstate `IKeysRepository` as a replacement for `NodeMasterKey`. (b) Give it some very-generic encrypt and sign methods which take as an argument the derivation path to the derived key to use. (c) Replace all other instances of `Key` in the code with a new type which wraps a reference to an `IKeysRepository` and a derivation path.

That way no secret data would ever leave the `IKeysRepository` type and a type which implements the `IKeysRepository` interface could store the key in mlocked memory or something.

Does that sound like a good way forward?

src/DotNetLightning.Core/Crypto/Keys.fs
canndrew: Fixed.

src/DotNetLightning.Core/Crypto/Keys.fs
canndrew: Fixed.

src/DotNetLightning.Core/Crypto/Keys.fs
canndrew: Fixed.

src/DotNetLightning.Core/Serialization/LightningStream.fs
canndrew: Fixed.

src/DotNetLightning.Core/Serialization/LightningStream.fs
canndrew: Fixed.

tests/DotNetLightning.Core.Tests/KeyRepositoryTests.fs
canndrew: Fixed.

src/DotNetLightning.Core/Chain/KeysInterface.fs
canndrew: I've the changed the commit title to use the word "refactor".

src/DotNetLightning.Core/Crypto/Keys.fs
canndrew: Fixed.

src/DotNetLightning.Core/Chain/KeysInterface.fs
canndrew: Fixed.

src/DotNetLightning.Core/Chain/KeysInterface.fs
canndrew: Fixed.

src/DotNetLightning.Core/Crypto/Keys.fs
canndrew: I've rewritten the commit message to more accurately describe what the commit is doing, including mentioning that it moves these methods across to the `ChannelPrivKeys` type.

src/DotNetLightning.Core/Channel/CommitmentsModule.fs
canndrew: Done.

knocte:
You added "Fixes issue 130" in the PR description but, did you add it to any commit? Which commit is the one that fixes the bug? It's hard to distinguish refactoring commits from the ones that change behaviour, for example, the commits whose title starts with "Add", do those change behaviour? Also the one that starts with "Reorder".

src/DotNetLightning.Core/Channel/Channel.fs
```diff
@@ -645,13 +647,14 @@ module Channel =
...
                 let cm = state.Commitments
+                let remoteChannelKeys = cm.RemoteParams.ChannelPubKeys
                 let lastCommitFeeSatoshi =
                     cm.FundingScriptCoin.TxOut.Value - (cm.LocalCommit.PublishableTxs.CommitTx.Value.TotalOut)
                 do! checkRemoteProposedHigherFeeThanBefore lastCommitFeeSatoshi msg.FeeSatoshis
                 let! closingTx, closingSignedMsg =
                     Closing.makeClosingTx (cs.KeysRepository, cm, state.LocalShutdown.ScriptPubKey, state.RemoteShutdown.ScriptPubKey, msg.FeeSatoshis, cm.LocalParams.ChannelPubKeys.FundingPubKey, cs.Network)
                     |> expectTransactionError
-                let! finalizedTx = Transactions.checkTxFinalized closingTx.Value (closingTx.WhichInput) (seq [ cm.RemoteParams.FundingPubKey, TransactionSignature(msg.Signature.Value, SigHash.All) ]) |> expectTransactionError
+                let! finalizedTx = Transactions.checkTxFinalized closingTx.Value (closingTx.WhichInput) (seq [ remoteChannelKeys.FundingPubKey, TransactionSignature(msg.Signature.Value, SigHash.All) ]) |> expectTransactionError
```
knocte: nit: let's split this line now that you're changing it, it's too long

src/DotNetLightning.Core/Channel/ChannelValidation.fs
```diff
@@ -90,45 +92,45 @@ module internal ChannelHelpers =
...
                                           (not localParams.IsFunder)
                                           (remoteParams.DustLimitSatoshis)
                                           revPubKeyForRemote
                                           localParams.ToSelfDelay
                                           delayedPubKeyForRemote
                                           paymentPubKeyForRemote
-                                          (remoteParams.HTLCBasePoint)
-                                          (localParams.ChannelPubKeys.HTLCBasePubKey)
+                                          (remoteChannelKeys.HTLCBasePubKey)
+                                          (localChannelKeys.HTLCBasePubKey)
```
knocte: nit: let's remove the parenthesis now that you're changing these lines, they are not needed

src/DotNetLightning.Core/Channel/CommitmentsModule.fs
```diff
@@ -356,7 +359,7 @@ module internal Commitments =
                         )
                     localHtlcSigsAndHTLCTxs |> List.map(fst), localHtlcSigsAndHTLCTxs |> List.map(snd) |> Seq.cast<IHTLCTx> |> List.ofSeq
 
-                let remoteHTLCPubKey = Generators.derivePubKey ctx (cm.RemoteParams.HTLCBasePoint) localPerCommitmentPoint.PubKey
+                let remoteHTLCPubKey = Generators.derivePubKey ctx (remoteChannelKeys.HTLCBasePubKey) localPerCommitmentPoint.PubKey
```
knocte: nit: let's remove the parenthesis now that you're changing this line, they are not needed

src/DotNetLightning.Core/Crypto/Keys.fs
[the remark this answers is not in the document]
knocte: then what needs to be done here is have the type be a discriminated union of only one element, and then there is no need to mark it with the `[<struct>]` attribute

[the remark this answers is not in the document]
knocte: (or a record, rather than a DU)

src/DotNetLightning.Core/Channel/ChannelValidation.fs
canndrew: This, and the other nit below, get fixed in a later commit anyway when these lines change again.

canndrew:
> You added "Fixes issue 130" in the PR description but, did you add it to any commit? Which commit is the one that fixes the bug? It's hard to distinguish refactoring commits from the ones that change behaviour, for example, the commits whose title starts with "Add", do those change behaviour? Also the one that starts with "Reorder".

The entire PR is refactoring. The bug was a naming issue where per-commitment-secrets were being called "revocation keys". This type gets renamed in one commit but an actual revocation key type isn't introduced until a later commit.

src/DotNetLightning.Core/Channel/Channel.fs
```diff
@@ -54,12 +62,20 @@ module Channel =
         |> fun ecdsaSig -> TransactionSignature(ecdsaSig.Value, SigHash.All)
 
     module Closing =
-        let makeClosingTx (keyRepo: IKeysRepository, cm: Commitments, localSpk: Script, remoteSpk: Script, closingFee: Money, localFundingPk, n) =
+        let makeClosingTx (
+                channelPrivKeys: ChannelPrivKeys,
+                cm: Commitments,
```
knocte, line 67: I think you're introducing indentation warnings here

src/DotNetLightning.Core/Channel/Channel.fs
```diff
@@ -16,25 +16,33 @@ open System
...
+    NodeSecret: NodeSecret
     State: ChannelState
     Network: Network
-    Secp256k1Context: ISecp256k1
  }
         with
-        static member Create(config, keysRepo, feeEstimator, localNodeSecret, fundingTxProvider, n, remoteNodeId) =
+        static member Create (
+                config: ChannelConfig,
+                nodeMasterPrivKey: NodeMasterPrivKey,
```
knocte, line 30: I think you're introducing indentation warnings here too

knocte: the way to fix I think is this way: instead of

```
Create (
    foo,
    bar
```

this way:

```
Create (foo,
             bar,
```

src/DotNetLightning.Core/Channel/Channel.fs
```diff
@@ -95,23 +119,24 @@ module Channel =
             [ MutualClosePerformed (closingTx, nextData) ]
             |> Ok
 
-        let claimCurrentLocalCommitTxOutputs (keyRepo: IKeysRepository, _channelPubKeys: ChannelPubKeys, commitments: Commitments, commitTx: CommitTx) =
+        let claimCurrentLocalCommitTxOutputs (
+                channelPrivKeys: ChannelPrivKeys,
+                commitments: Commitments,
+                commitTx: CommitTx
```
knocte, line 125: same here (indentation warnings)

src/DotNetLightning.Core/Channel/ChannelValidation.fs
```diff
@@ -13,26 +13,33 @@ open DotNetLightning.Transactions
...
 
     let private makeFlags (isNode1: bool, enable: bool) =
         (if isNode1 then 1uy else 0uy) ||| ((if enable then 1uy else 0uy) <<< 1)
 
-    let internal makeChannelUpdate (chainHash, nodeSecret: Key, remoteNodeId: NodeId, shortChannelId, cltvExpiryDelta,
-                                    htlcMinimum, feeBase, feeProportionalMillionths, enabled: bool, timestamp) =
+    let internal makeChannelUpdate (
+            chainHash,
+            nodeSecret: NodeSecret,
+            remoteNodeId: NodeId,
```
knocte, line 32: same here (indentation warnings)

src/DotNetLightning.Core/Crypto/Keys.fs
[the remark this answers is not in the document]
joemphilips: I agree with @knocte , DU (or record. I prefer DU though) achieves the same.
Using struct is possible in every case. (i.e. DU, or Record, or plain old class.). But it makes harder to reason about the behavior. the only case I think it is feasible is that when the inner value is also a value type and thus we can gain performance by preventing heap-allocation.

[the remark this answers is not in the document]
knocte: I would go as far as to say that idiomatic F# only needs records and DUs, types (that end up compiled in the same way as C# classes) should only be used when trying to achieve interoperability with C#

src/DotNetLightning.Core/Chain/KeysInterface.fs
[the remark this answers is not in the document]
joemphilips: I think you understand what I meant perfectly (Indeed better than me.)

> Firstly, IKeysRepository shouldn't have an option to get the channel private keys at all, otherwise we'll end up holding them in RAM anyway. Even if the code only grabs the keys temporarily when it needs them we don't have a way to control when those keys get garbage collected. Also they'll still live on in RAM for a while after they get collected, assuming the CLR doesn't zero-out memory once it gets collected since that would be an expensive waste of time when the memory is being used for anything other than holding secrets.

Yeah this is very good point. I had to make `ChannelKeys` a `ref struct`. I actually tried once, but it does not go well with functional style. computation expression did not work with ref struct.

> So maybe a good design here would be to (a) reinstate IKeysRepository as a replacement for NodeMasterKey. (b) Give it some very-generic encrypt and sign methods which take as an argument the derivation path to the derived key to use. (c) Replace all other instances of Key in the code with a new type which wraps a reference to an IKeysRepository and a derivation path.

This sounds good. But I think the refactoring will be very hard. Unless you want to try, I think we should go as is so that we can first get shit done.

So yeah, I'm not ok with removing the `IKeysRepository` there are tons of other things we have to consider in terms of security anyway...

src/DotNetLightning.Core/Serialization/LightningStream.fs
```diff
@@ -325,20 +325,22 @@ type LightningReaderStream(inner: Stream) =
         let len = this.ReadUInt16(false)
         this.ReadBytes(int32 len)
 
+    member this.ReadKey() =
```
knocte, line 328: should this rather be called ReadPrivKey()?

src/DotNetLightning.Core/Crypto/Keys.fs
knocte: this still has `and`, so not fixed actually

src/DotNetLightning.Core/Chain/KeysInterface.fs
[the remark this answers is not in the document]
knocte: @canndrew please note that it seems Joe edited his last comment to change `not` to `now` hehe ;-)

src/DotNetLightning.Core/Channel/Channel.fs
canndrew: Fixed.

src/DotNetLightning.Core/Channel/Channel.fs
canndrew: Fixed.

src/DotNetLightning.Core/Serialization/LightningStream.fs
canndrew: I named it `ReadKey` because the NBitcoin type (which it returns) is named `Key`. Though I agree the lack of consistency between "Key" and "PrivKey" is bad.

I've added a type annotation to the method (and all the other methods of `LightningStream`) to make this slightly clearer.

src/DotNetLightning.Core/Channel/ChannelValidation.fs
canndrew: Fixed.

src/DotNetLightning.Core/Channel/Channel.fs
canndrew: Fixed.

src/DotNetLightning.Core/Channel/Channel.fs
canndrew: Fixed.

src/DotNetLightning.Core/Channel/Channel.fs
canndrew: Fixed.

src/DotNetLightning.Core/Crypto/Keys.fs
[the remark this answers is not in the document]
canndrew: I've changed the key types to be single-constructor discriminated unions.

src/DotNetLightning.Core/Crypto/Keys.fs
```diff
@@ -0,0 +1,662 @@
+namespace DotNetLightning.Crypto
+
+open System
+open NBitcoin
+open NBitcoin.Crypto
+
+open DotNetLightning.Utils
+open DotNetLightning.Core.Utils.Extensions
+
+// FIXME: Should the [<Struct>]-annotated types here be changed to records or single-constructor discriminated unions?
```
knocte, line 10: does this comment apply anymore?

src/DotNetLightning.Core/Crypto/Keys.fs
```diff
@@ -0,0 +1,662 @@
...
+
+    let Secp256k1 = CryptoUtils.impl.newSecp256k1()
+
+    type Key with
+        static member Mul(lhs: Key, rhs: Key): Key =
+            let lhsBytes = lhs.ToBytes()
+            let rhsBytes = rhs.ToBytes()
+            let tweak = ReadOnlySpan(lhsBytes)
+            match Secp256k1.PrivateKeyTweakMultiply(tweak, rhsBytes.AsSpan()) with
+            | true -> Key rhsBytes
```
knocte, line 28: should be new Key(rhsBytes) to avoid warning?

src/DotNetLightning.Core/Crypto/Keys.fs
```diff
@@ -0,0 +1,662 @@
...
+            | false -> failwith "failed to multiply Keys"
+
+        static member Add(lhs: Key, rhs: Key): Key =
+            let lhsBytes = lhs.ToBytes()
+            let rhsBytes = rhs.ToBytes()
+            let tweak = ReadOnlySpan(lhsBytes)
+            match Secp256k1.PrivateKeyTweakAdd(tweak, rhsBytes.AsSpan()) with
+            | true -> Key rhsBytes
```
knocte, line 36: same here

src/DotNetLightning.Core/Crypto/Keys.fs
canndrew: Fixed.

src/DotNetLightning.Core/Crypto/Keys.fs
canndrew: Fixed.

src/DotNetLightning.Core/Crypto/Keys.fs
canndrew: Nope. Fixed.

src/DotNetLightning.Core/Utils/NBitcoinExtensions.fs
```diff
@@ -35,6 +36,9 @@ module NBitcoinExtensions =
     type Key with
         static member BytesLength: int = 32
 
+        static member FromHash(preimage: array<byte>): Key =
+            Key (Hashes.SHA256 preimage)
```
knocte, line 40: argh another `new` keyword missing, sorry for not spotting it earlier

canndrew: Ditto. However we really need to fix all the build warnings so that these things are easier to spot.

Fixed.

knocte:
LGTM, @joemphilips ?

knocte:
Wait! @canndrew the commit "Reorder definitions in Crypto.Keys" failed CI :(

src/DotNetLightning.Core/Utils/NBitcoinExtensions.fs
knocte: well, I tried in one of my latest PRs, there was a warning that I didn't manage to get rid of... beyond my paygrade.

canndrew:
Hmmm. A random failure from an unstable test maybe? The commits following it passed and they only contain trivial formatting changes.

canndrew:
I just ran the tests for the commit 10 times on my machine and it didn't fail...

knocte:
Maybe a test affected by https://github.com/joemphilips/DotNetLightning/issues/99?

joemphilips:
I've once had similar issue and ending up finding that printing something into stdout with printfn caused to crash Expecto. (https://github.com/haf/expecto#about-test-parallelism)
 Maybe that is the case here?

joemphilips:
It seems that test for Macaroon is failing in some case, I will look into it later. You can just ignore now.

src/DotNetLightning.Core/Crypto/Keys.fs
```diff
@@ -0,0 +1,660 @@
...
+        PaymentPubKey =
+            this.DerivePaymentPubKey channelPubKeys.PaymentBasepoint
+        DelayedPaymentPubKey =
+            this.DeriveDelayedPaymentPubKey channelPubKeys.DelayedPaymentBasepoint
+        HtlcPubKey =
+            this.DeriveHtlcPubKey channelPubKeys.HtlcBasepoint
+    }
+
+type CommitmentNumber =
+    | CommitmentNumber of UInt48
```
joemphilips, line 390: `Uint48` is a value type (and its inner value is value type too) so it is good to make this type struct too.

src/DotNetLightning.Core/DotNetLightning.Core.fsproj
```diff
@@ -40,6 +40,12 @@
     <Compile Include="Utils/Errors.fs" />
     <Compile Include="Utils/RouteType.fs" />
     <Compile Include="Utils/PriorityQueue.fs" />
+    <Compile Include="Crypto/CryptoUtils.fs" />
```
joemphilips, line 43: Why you had to move these files? I want the file order in fsproj to be sorted with directory first.

src/DotNetLightning.Core/Crypto/Keys.fs
```diff
@@ -0,0 +1,660 @@
...
+                        (remotePaymentBasepoint: PaymentBasepoint)
+                            : ObscuredCommitmentNumber =
+        let obscureFactor =
+            CommitmentNumber.ObscureFactor
+                isFunder
+                localPaymentBasepoint
+                remotePaymentBasepoint
+        ObscuredCommitmentNumber((UInt48.MaxValue - this.Index()) ^^^ obscureFactor)
+
+and ObscuredCommitmentNumber =
```
joemphilips, line 451: So as this.

src/DotNetLightning.Core/Utils/NBitcoinExtensions.fs
```diff
@@ -35,6 +36,9 @@ module NBitcoinExtensions =
     type Key with
         static member BytesLength: int = 32
 
+        static member FromHash(preimage: array<byte>): Key =
```
joemphilips, line 39: The name is confusing, it sounds like the argument is hash itself.
I suggest `FromHashOf` instead.

src/DotNetLightning.Core/DotNetLightning.Core.fsproj
joemphilips: I mean, i don't want to have something like.

```fsproj
<Compile Include="A/Foo.fs" />
<Compile Include="B/Bar.fs" />
<Compile Include="A/Buz.fs" /> <!-- A has been splitted into two! -->
```

For the sake of clarity and ease of understanding.
If this is really necessary, then we should introduce another repository.

canndrew: It's no clear to me what the best way to break the files into separate dirs would be. Aezeed is closely related to keys, but it depends on the serialization types (`BitWriter` and `LightningWriterStream`), these types (specifically `LightningReaderStream`) depend on the key types though.

Perhaps the serialization methods should be moved from `LightningReaderStream` onto the key types as an interface? eg. have the key types implement the `ILightningSerializable` interface, then make `Lightning{Reader,Writer}Stream` be able to read/write any `ILightningSerializable` type?

joemphilips: Why don't you just put new key types in `Utils/` and extend its methods in `Crypto/` ? I think that is more easier to reason about which logic is in which folder.

joemphilips: To rephrase my thinking for the sake of clarity....

*The entire `Serialization/` folder should work without any cryptographic stuff. So putting `Crypto/` above `Serialization/` must be unnecessary and responsibility gets less segregated.
* Why you want to do this is because (I guess) you want to use new Key types as both 1. Primitive building block for serialization (like LNMoney) and 2. entrypoint for cryptographic operation (e.g. key tweaking). But these two characteristics are logically in different layer. So crypto must come after `Serialization/` and if you want new key types to have cryptographic methods, it must be in the form of extension methods in `Crypto/`

canndrew: > Why don't you just put new key types in `Utils/` and extend its methods in `Crypto/` ?

Because I didn't think of that :)

I've done that now.

src/DotNetLightning.Core/Utils/NBitcoinExtensions.fs
canndrew: Fixed.

src/DotNetLightning.Core/Crypto/Keys.fs
canndrew: Fixed.

src/DotNetLightning.Core/Crypto/Keys.fs
canndrew: Fixed.

canndrew:
I've addressed the feedback and rebased this off of master.

knocte:
The CI of commit "Add ReadKey method and avoid magic number in LightningStream" has failed but it seems it's something that can be fixed by rerunning it, can you retry CI?

src/DotNetLightning.Core/Utils/Keys.fs
```diff
@@ -0,0 +1,348 @@
...
+    static member FirstCommitment: CommitmentNumber =
+        CommitmentNumber UInt48.MaxValue
+
+    member this.PreviousCommitment(): CommitmentNumber =
+        CommitmentNumber(this.Index() + UInt48.One)
+
+    member this.NextCommitment(): CommitmentNumber =
+        CommitmentNumber(this.Index() - UInt48.One)
+
+and [<Struct>] ObscuredCommitmentNumber =
```
knocte, line 279: @canndrew does this really need an `and`? doesn't look like

canndrew: Fixed.

joemphilips (approved):

PR #137, opened 2020-10-28
Workaround NBitcoin txout shuffling bug

canndrew:
NBitcoin's `TransactionBuilder.ShuffleOutputs` flag is broken and gets ignored when building the `Transaction`. This can be worked around by setting the `TransactionBuilder`'s `RandomShuffle` field to `null`, which disables all random shuffling.

This fix can be removed once NBitcoin bug #931 is fixed:

See:
https://github.com/MetacoSA/NBitcoin/issues/931


joemphilips:
LGTM

PR #138, opened 2020-10-28
Add TxOut.LexicographicCompare method

canndrew:
Ordering txouts by their byte representation doesn't give a valid lexicographic ordering due to (a) the amount being represented in little-endian (BIP69 requires that txouts with smaller values appear first) and (b) F#'s comparision function on byte arrays always ordering smaller arrays first (eg.` [| 1 |]` < `[| 0; 0 |]` < `[| 1; 1 |]`), rather than ordering them in dictionary order (ie.` [| 0; 0 |]` < `[| 1 |]` < `[| 1; 1 |]`).

This commit adds a `TxOut.LexicographicCompare` method which correctly implements BIP69's txout lexicographic ordering spec. This is then used for ordering outputs in the closing transaction.

### DIFF
--- a/src/DotNetLightning.Core/Utils/NBitcoinExtensions.fs
+++ b/src/DotNetLightning.Core/Utils/NBitcoinExtensions.fs
@@ -22,6 +22,32 @@ module NBitcoinExtensions =
             res.[31] <- res.[31] ^^^ (uint8 (this.N >>> 0) &&& 0xffuy)
             res  |> uint256 |> ChannelId
 
+    type TxOut with
+        static member LexicographicCompare (txOut0: TxOut)
+                                           (txOut1: TxOut)
+                                               : int =
+            if txOut0.Value < txOut1.Value then
+                -1
+            elif txOut0.Value > txOut1.Value then
+                1
+            else
+                let script0 = txOut0.ScriptPubKey.ToBytes()
+                let script1 = txOut1.ScriptPubKey.ToBytes()
+                let rec compare (index: int) =
+                    if script0.Length = index && script1.Length = index then
+                        0
+                    elif script0.Length = index then
+                        -1
+                    elif script1.Length = index then
+                        1
+                    elif script0.[index] < script1.[index] then
+                        -1
+                    elif script0.[index] > script1.[index] then
+                        1
+                    else
+                        compare (index + 1)
+                compare 0
+
     type PSBT with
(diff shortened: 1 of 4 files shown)


knocte:
Good work!

This looks like the perfect case in which we should have tests cause they should be easy to do (no cross boundary, just plain *unit* test AFAIU).

joemphilips:
Interesting, thanks for reporting.
I agree with @knocte that this needs test.

joemphilips:
bip69 test vector must be sufficient.

src/DotNetLightning.Core/Utils/NBitcoinExtensions.fs
```diff
@@ -22,6 +22,32 @@ module NBitcoinExtensions =
...
+                                               : int =
+            if txOut0.Value < txOut1.Value then
+                -1
+            else if txOut0.Value > txOut1.Value then
+                1
+            else
+                let script0 = txOut0.ScriptPubKey.ToBytes()
+                let script1 = txOut1.ScriptPubKey.ToBytes()
+                let rec compare (index: int) =
+                    if (script0.Length = index) && (script1.Length = index) then
```
knocte, line 37: nit: no need for parens here

src/DotNetLightning.Core/Utils/NBitcoinExtensions.fs
```diff
@@ -22,6 +22,32 @@ module NBitcoinExtensions =
...
+                        0
+                    else if script0.Length = index then
+                        -1
+                    else if script1.Length = index then
+                        1
+                    else if script0.[index] < script1.[index] then
+                        -1
+                    else if script0.[index] > script1.[index] then
```
knocte, line 45: nit: you can replace all `else if` occurrences with `elif`

canndrew: Fixed.

src/DotNetLightning.Core/Utils/NBitcoinExtensions.fs
canndrew: Fixed.

canndrew:
I've added another commit which adds tests. The examples in BIP69 are pretty insufficient for testing so I wrote my own test cases to test all the possible combinations of relevant differences.

knocte:
LGTM (nit: tests are introducing new warnings)

PR #143, opened 2021-02-02
Prevent possible F#4.0 issue

knocte:
It seems that these attributes can cause this compilation error in F#4.0:
error FS0927: The kind of the type specified by its attributes does not match the kind implied by its definition

Newer versions of F# (like 4.5 or even newer, like the one being used
by .NETCore to build the binary that is later published in nuget) allow
compiling this code with no issues, but if you reference the generated
assembly later from an old F# compiler, it could generate exceptions at
runtime, e.g.: System.BadFormatImageException (or other types) whose
inner exception could be the following:

System.TypeLoadException : Could not load type of field 'GWallet.Backend.UtxoCoin.Lightning.SerializedChannel:MinSafeDepth@' (6) due to: Expected reference type but got type kind 17

FSharp.Core's Result type is also affected by this so in this commit we
create a replacement for it that is only used in the BouncyCastle build
(which we now rename as 'Portability' build).

This commit can be reverted once it is assumed that all consumers of
DotNetLightning can use at least F# 4.5. To make sure this happens
even in distros where no Microsoft’s binary blobs are consumed (e.g.
Debian and Ubuntu without using Microsoft apt repositories) we have
proposed an upgrade of the F# toolset downstream, which has been
merged and is in the process of being adopted:
https://salsa.debian.org/dotnet-team/fsharp/-/merge_requests/2

Co-authored-by: Andres G. Aragoneses <knocte@gmail.com>
Co-authored-by: Andrew Cann <shum@canndrew.org>

### DIFF
--- a/src/ResultUtils/Result.fs
+++ b/src/ResultUtils/Result.fs
@@ -1,7 +1,29 @@
+namespace ResultUtils.Portability
+
+#if NoDUsAsStructs
+[<StructuralEquality; StructuralComparison>]
+[<CompiledName("FSharpResult`2")>]
+type Result<'T,'TError> =
+    | Ok of ResultValue: 'T
+    | Error of ErrorValue: 'TError
+#endif
+
+
 namespace ResultUtils
+
+open ResultUtils.Portability
 
 [<RequireQualifiedAccess>]
 module Result =
+
+  let ToFSharpCoreResult res =
+#if NoDUsAsStructs
+    match res with
+    | Ok o -> FSharp.Core.Result.Ok o
+    | Error e -> FSharp.Core.Result.Error e
+#else
+    res
+#endif
 
   let isOk x =
     match x with
@@ -17,11 +39,20 @@ module Result =
     | Error err -> errorF err
 
   let eitherMap okF errorF x =
-    either (okF >> Result.Ok) (errorF >> Result.Error) x
+    either (okF >> Ok) (errorF >> Error) x
+
+  let bind binder result =
+    match result with Error e -> Error e | Ok x -> binder x
+
+  let map mapping result =
+    match result with Error e -> Error e | Ok x -> Ok (mapping x)
+
+  let mapError mapping result =
+    match result with Error x -> Error (mapping x) | Ok v -> Ok v
 
   let apply f x =
-    Result.bind (fun f' ->
-      Result.bind (fun x' -> Ok (f' x')) x) f
+    bind (fun f' ->
+      bind (fun x' -> Ok (f' x')) x) f
 
   let map2 f x y =
     (apply (apply (Ok f) x) y)
@@ -43,11 +74,11 @@ module Result =
     let tryCreate' x =
       (^b : (static member TryCreate : 'a -> Result< ^b, 'c>) x)
     tryCreate' x
-    |> Result.mapError (fun z -> (fieldName, z))
+    |> mapError (fun z -> (fieldName, z))
 
   /// Replaces the wrapped value with unit
   let ignore result =
-    result |> Result.map ignore
+    result |> map ignore
 
   /// Returns the specified error if the value is false.
   let requireTrue error value =
@@ -104,12 +135,12 @@ module Result =
 
   /// Replaces an error value with a custom error value.
   let setError error result =
-    result |> Result.mapError (fun _ -> error)
+    result |> mapError (fun _ -> error)
 
   /// Replaces a unit error value with a custom error value. Safer than setError
   /// since you're not losing any information.
   let withError error result =
-    result |> Result.mapError (fun () -> error)
+    result |> mapError (fun () -> error)
 
   /// Returns the contained value if Ok, otherwise returns ifError.
   let defaultValue ifError result =
--- a/tests/DotNetLightning.Core.Tests/PaymentTests.fs
+++ b/tests/DotNetLightning.Core.Tests/PaymentTests.fs
@@ -6,10 +6,12 @@ open DotNetLightning.Payment
 open DotNetLightning.Utils
 
 open DotNetLightning.Serialization
-open ResultUtils
 open Expecto
 open NBitcoin
 open NBitcoin.Crypto
+
+open ResultUtils
+open ResultUtils.Portability
 
 [<Tests>]
 let tests =
@@ -184,7 +186,7 @@ let tests =
         testCase "Same, but adding invalid unknown feature 100" <| fun _ ->
             let data = "lnbc25m1pvjluezpp5qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypqdq5vdhkven9v5sxyetpdeessp5zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zygs9q4psqqqqqqqqqqqqqqqpqsqq40wa3khl49yue3zsgm26jrepqr2eghqlx86rttutve3ugd05em86nsefzh4pfurpd9ek9w2vp95zxqnfe2u7ckudyahsa52q66tgzcp6t2dyk"
             let pr = PaymentRequest.Parse(data)
-            Expect.isError(pr) ""
+            Expect.isError (Result.ToFSharpCoreResult  pr) ""
     ]
     
 [<Tests>]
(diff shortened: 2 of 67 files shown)


tests/DotNetLightning.Core.Tests/PaymentTests.fs
```diff
@@ -184,7 +186,7 @@ let tests =
         testCase "Same, but adding invalid unknown feature 100" <| fun _ ->
             let data = "lnbc25m1pvjluezpp5qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypqdq5vdhkven9v5sxyetpdeessp5zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zygs9q4psqqqqqqqqqqqqqqqpqsqq40wa3khl49yue3zsgm26jrepqr2eghqlx86rttutve3ugd05em86nsefzh4pfurpd9ek9w2vp95zxqnfe2u7ckudyahsa52q66tgzcp6t2dyk"
             let pr = PaymentRequest.Parse(data)
-            Expect.isError(pr) ""
+            Expect.isError (Result.ToFSharpCoreResult  pr) ""
```
joemphilips, line 189: Is it necessary to do this in tests? I think supporting tests only for environments listed in the CI is enough. And there was no problem so far in the CI.

canndrew: `PaymentRequest.Parse` returns a `ResultUtils.Portability.Result` but `Expect.isError` takes an `FSharp.Core.Result`. So there needs to be a conversion here.

Or do you mean that the tests should be able to assume that DNL was built without the `NoDUsAsStructs` flag? If so, I think we may as well make the tests compatible with old F# if we're making everything else compatible with old F#, since the changes aren't very invasive (just lots of calls to `Result.ToFSharpCoreResult` and lots of `open ResultUtils.Portability`). Perhaps we could add a helper function that wraps `Result.ToFSharpCoreResult >> Expect.isError`.

canndrew: Of course, having to make these changes at all is pretty unfortunate. But it's pretty much a necessity until Ubuntu et. al. update to a non-ancient F# compiler. When they do we can revert these changes.

src/ResultUtils/Result.fs
```diff
@@ -1,8 +1,30 @@
+namespace ResultUtils.Portability
```
joemphilips, line 1: I think 1 namespace per 1 file is more easy to understand.
Also, do you really need this namespace? it seems writing `open ResultUtils.Portability` in may files seems unnecessary change.

joemphilips: Why don't we create another file and define the Result type in there, with the namespace `FSharp.Core.Result`?

tests/DotNetLightning.Core.Tests/PaymentTests.fs
joemphilips: > Or do you mean that the tests should be able to assume that DNL was built without the NoDUsAsStructs flag?

No this is not what meant, tests should be done in both builds.

> PaymentRequest.Parse returns a ResultUtils.Portability.Result but Expect.isError takes an FSharp.Core.Result. So there needs to be a conversion here.

Wait than you mean when we revert this change, it will be a breaking change for the consumer.
(since the `Result` will be a different type(`FSharp.Core.Result` -> `ResultUtils.Portability.Result`), right?)
I don't want to do this ... why can we just write in the readme to install the latest dotnet core sdk or mono before using it?

knocte: > why can we just write in the readme to install the latest dotnet core sdk or mono before using it?

Because that would mean trusting Microsoft, because the only way to install F# is installing their binaries. Do you trust Microsoft with your BTC funds? I kinda do, but as a wallet developers, I don't want my users to have to trust Microsoft. This is the reason my team has contributed the F#4.5 update to Debian: to have a way to install F#4.5 without using Microsoft binaries.

> it will be a breaking change for the consumer.

For the consumer of DNL yes. Do you many consumers? I think there's only geewallet now. And by the time geewallet becomes production ready (or any other wallet that might use DNL in the future), I'm sure F#4.5 would have landed in Debian already and we would have reverted this hack already.

knocte: (edited the above comment for clarity)

joemphilips: > Because that would mean trusting Microsoft, because the only way to install F#4.5 is installing their binaries. Do you trust Microsoft with your BTC funds? I kinda do, but as a wallet developer, I don't want my users to have to trust Microsoft. 

Everything is OSS, if users don't want to trust MS binaries, they can always  compile F#4.5 from the source code. So I think this argument does not hold here.

> For the consumer of DNL yes. Do you know many consumers? I think there's only geewallet now.

I agree there are not much consumers. But I myself am in https://github.com/joemphilips/NRustLightning

Let's just assume that users has a F# 4.5 on their environment. I think that way is more simple.

knocte: > Everything is OSS, if users don't want to trust MS binaries, they can always compile F#4.5 from the source code. So I think this argument does not hold here.

And you think someone that is interested in installing geewallet but run it in a trustless way is going to go ahead with all the effort that requires compiling .NET5 & F# from scratch? That's the point of using Debian.

> I agree there are not much consumers. But I myself am in https://github.com/joemphilips/NRustLightning

But you're using the NSec build right? So this doesn't affect you.

knocte: PS: I'm not a debian dev/user myself, but contributing to it is the only way to make packages land in Ubuntu, at least the .NET related ones. Having an Ubuntu distro with .NET installed from Canonical binary repositories gives me (and many others) much more confidence than trusting any ISV, plus is user-friendly enough to not having to compile everything (we know how Gentoo ended up in terms of market-share).

joemphilips: > And you think someone that is interested in installing geewallet but run it in a trustless way is going to go ahead with all the effort that requires compiling .NET5 & F# from scratch?

Actually, yes. it is the cost for trustless-ness.

The point you are making is that users should blindly trust binary in Debian canonical repository rather than blindly trusting MS binary.
So I guess the point is not in the trustless-ness but how easy for users to install F#4.5.
I don't fully agree that installing F# from canonical repo for the distro is much easier than installing binary from the MS. (That is what linux users regularly do when they can not find the software in the canonical repo.)
So I don't think the pro(the easiness of installation) exceeds con (the breaking change.) in this case.

But I don't know, since it is true that you are the primary consumer of the BouncyCastle build of this package, I can merge it if you really want it. But I want you to revert this change when the latest debian/ubuntu gets ready for it, is that OK?

> But you're using the NSec build right? So this doesn't affect you.

Yeah that's right. At least for now.

knocte: > So I guess the point is not in the trustless-ness but how easy for users to install F#4.5.

No. I mean, sure, it's much easier to install fsharp by just typing `apt install fsharp` than visiting a website from MS to extract some script that you execute to add their apt repositories and install the packages, but no, my main point is that trusting Debian is much more logical than trusting Microsoft. When you trust Debian you're trusting a whole infrastructure system of community peer review and meritocracy (because only the most experienced people can become Debian Developers). Not only that, Debian has huge efforts towards reproducible builds (e.g. the whole point of their recent infrastructure efforts with salsa.debian.org integrating GitLabCI is so that maintainers don't have to build on their boxes anymore, so instead of uploading binaries they upload sources). I think Microsoft is doing some stuff around reproducible builds but they're not quite there yet.

>  But I want you to revert this change when the latest debian/ubuntu gets ready for it, is that OK?

Of course, it's us who are proposing to reverting this hack. In fact, this is why we spent so much time upgrading F# 4.5 on debian (have you seen the Merge Request? it's huge). Because otherwise we cannot even consume FSharp.Core's Result type, which is a thing that is being pushed by the F# community recently (to stop relying on impure exceptions so much).

We will address the namespace issue and ping you when we have updated the PR.

joemphilips: > my main point is that trusting Debian is much more logical than trusting Microsoft. When you trust Debian you're trusting a whole infrastructure system of community peer review and meritocracy (because only the most experienced people can become Debian Developers). Not only that, Debian has huge efforts towards reproducible builds (e.g. the whole point of their recent infrastructure efforts with salsa.debian.org integrating GitLabCI is so that maintainers don't have to build on their boxes anymore, so instead of uploading binaries they upload sources). I think Microsoft is doing some stuff around reproducible builds but they're not quite there yet.

Great, I see, I'm now convinced.

src/ResultUtils/Result.fs
canndrew: Using the namespace `FSharp.Core.Result` seems like a recipe for headaches since there's already an `FSharp.Core.Result` namespace with a `Result` type. It'll make it hard to disambiguate between the two `Result` types in the same namespace and to make sure you're using the one you intend to. And It'll make the code and the inevitable compilation errors a lot more confusing.

joemphilips: > Using the namespace FSharp.Core.Result seems like a recipe for headaches since there's already an FSharp.Core.Result namespace with a Result type. It'll make it hard to disambiguate between the two Result types in the same namespace and to make sure you're using the one you intend to. And It'll make the code and the inevitable compilation errors a lot more confusing.

Right, this is fair point.

What about my first comment?

canndrew: I can move the namespace to another file easily enough. I'll do a PR to this PR for that.

PR #144, opened 2021-02-10
Fix fee rate calculation

knocte:
FeeRatePerKw.FromFee was mis-calculating the fee rate from the fee and weight.

By @canndrew


knocte:
@joemphilips actually I'm surprised you merged this so fast (without asking us to write tests)?

joemphilips:
Well the mistake seemed obvious.
I would like you to write a test though.
(The root cause of these bugs are that we still don't have tests for Channel state management. I'm not planning to do it with myself since I don't intend to use that part.)

PR #146, opened 2021-02-10
Replace ProvideFundingTx with WaitForFundingTx state

canndrew:
For some of the work we're doing on geewallet it became problematic for us to have to pass in a `ProvideFundingTx` when initially creating a `Channel`.

This commit changes things so that when creating a channel, instead of passing DNL a closure which takes the funding tx parameters and returns the funding tx, DNL will now returns the funding tx parameters and requires us to pass it the funding tx using a new `CreateFundingTx` command.

This is also has the extra type-safety advantage of us not having to pass DNL a dummy `ProvideFundingTx` (which just calls `failwith`) when creating an inbound channel.

### DIFF
--- a/src/DotNetLightning.Core/Channel/Channel.fs
+++ b/src/DotNetLightning.Core/Channel/Channel.fs
@@ -13,12 +13,10 @@ open System
 open ResultUtils
 open ResultUtils.Portability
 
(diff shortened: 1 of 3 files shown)


joemphilips:
LGTM, I will rebase and merge.

knocte:
@joemphilips next time you rebase, can you use `git rebase` instead of `git merge`? I think it's better to have a linear git history.

PR #148, opened 2021-03-16
Add support for force-closing channels

canndrew:
This PR adds support for force-closing channels.

In order to support this, this PR adds three things:

1) `CommitmentToLocalExtension`

This is an `NBitcoin` `BuilderExtension` that tells `NBitcoin.TransactionBuilder` how to recognize and sign `to_local` txouts from lightning commitment transactions. This is needed for force-closing to spend `to_local` outputs.

2) `tryGetFundsFromLocalCommitmentTx`

When force-closing a channel this function is used to recover funds from our commitment transaction. It returns a `TransactionBuilder` which is used to generate a recovery transaction which can be broadcast once the timelock on the `to_local` output has expired.

3) Two new computation expressions

`OptionCE` is a computation expression for creating options, similar to the result computation expression which DNL already has. `SeqConsumerCE` is a computation expression which makes it easy to write code which consumes a sequence, one element at a time.

The upgrade NBitcoin to v5.0.75 is done so that the function `tryGetFundsFromLocalCommitmentTx` can use the new sequence number and version `TransactionBuilder` setting capabilities.

### DIFF
--- a/src/DotNetLightning.Core/Channel/CommitmentToLocalExtension.fs
+++ b/src/DotNetLightning.Core/Channel/CommitmentToLocalExtension.fs
@@ -1,0 +1,130 @@
+namespace DotNetLightning.Channel
+
+open System
+open NBitcoin
+open NBitcoin.BuilderExtensions
+open DotNetLightning.Utils
+open DotNetLightning.Crypto
+
+open ResultUtils
+open ResultUtils.Portability
+
+type CommitmentToLocalParameters = {
+    ToSelfDelay: BlockHeightOffset16
+    LocalDelayedPubKey: DelayedPaymentPubKey
+}
+    with
+    static member TryExtractParameters (scriptPubKey: Script): Option<CommitmentToLocalParameters> =
+        let ops =
+            scriptPubKey.ToOps()
+            // we have to collect it into a list and convert back to a seq
+            // because the IEnumerable that NBitcoin gives us is internally
+            // mutable.
+            |> List.ofSeq
+            |> Seq.ofList
+        let checkOpCode(opcodeType: OpcodeType) = seqParser<Op> {
+            let! op = SeqParser.next()
+            if op.Code = opcodeType then
+                return ()
+            else
+                return! SeqParser.abort()
+        }
+        let parseToCompletionResult =
+            SeqParser.parseToCompletion ops <| seqParser {
+                do! checkOpCode OpcodeType.OP_IF
+                let! opRevocationPubKey = SeqParser.next()
+                let! _revocationPubKey = seqParser {
+                    match opRevocationPubKey.PushData with
+                    | null -> return! SeqParser.abort()
+                    | bytes ->
+                        try
+                            return RevocationPubKey.FromBytes bytes
+                        with
+                        | :? FormatException -> return! SeqParser.abort()
+                }
+                do! checkOpCode OpcodeType.OP_ELSE
+                let! opToSelfDelay = SeqParser.next()
+                let! toSelfDelay = seqParser {
+                    let nullableToSelfDelay = opToSelfDelay.GetLong()
+                    if nullableToSelfDelay.HasValue then
+                        try
+                            return BlockHeightOffset16 (Convert.ToUInt16 nullableToSelfDelay.Value)
+                        with
+                        | :? OverflowException -> return! SeqParser.abort()
+                    else
+                        return! SeqParser.abort()
+                }
+                do! checkOpCode OpcodeType.OP_CHECKSEQUENCEVERIFY
+                do! checkOpCode OpcodeType.OP_DROP
+                let! opLocalDelayedPubKey = SeqParser.next()
+                let! localDelayedPubKey = seqParser {
+                    match opLocalDelayedPubKey.PushData with
+                    | null -> return! SeqParser.abort()
+                    | bytes ->
+                        try
+                            return DelayedPaymentPubKey.FromBytes bytes
+                        with
+                        | :? FormatException -> return! SeqParser.abort()
+                }
+                do! checkOpCode OpcodeType.OP_ENDIF
+                do! checkOpCode OpcodeType.OP_CHECKSIG
+                return {
+                    ToSelfDelay = toSelfDelay
+                    LocalDelayedPubKey = localDelayedPubKey
+                }
+            }
+        match parseToCompletionResult with
+        | Ok data -> Some data
+        | Error _consumeAllError -> None
+
+type internal CommitmentToLocalExtension() =
+    inherit BuilderExtension()
+        override self.CanGenerateScriptSig (scriptPubKey: Script): bool =
+            (CommitmentToLocalParameters.TryExtractParameters scriptPubKey).IsSome
+
+        override self.GenerateScriptSig(scriptPubKey: Script, keyRepo: IKeyRepository, signer: ISigner): Script =
+            let parameters =
+                match (CommitmentToLocalParameters.TryExtractParameters scriptPubKey) with
+                | Some parameters -> parameters
+                | None ->
+                    failwith
+                        "NBitcoin should not call this unless CanGenerateScriptSig returns true"
+            let pubKey = keyRepo.FindKey scriptPubKey
+            // FindKey will return null if it can't find a key for
+            // scriptPubKey. If we can't find a valid key then this method
+            // should return null, indicating to NBitcoin that the sigScript
+            // could not be generated.
+            match pubKey with
+            | null -> null
+            | _ when pubKey = parameters.LocalDelayedPubKey.RawPubKey() ->
+                let localDelayedSig = signer.Sign (parameters.LocalDelayedPubKey.RawPubKey())
+                Script [
+                    Op.GetPushOp (localDelayedSig.ToBytes())
+                    Op.op_Implicit OpcodeType.OP_FALSE
+                ]
+            | _ -> null
+
+        override self.CanDeduceScriptPubKey(_scriptSig: Script): bool =
+            false
+
+        override self.DeduceScriptPubKey(_scriptSig: Script): Script =
+            raise <| NotSupportedException()
+
+        override self.CanEstimateScriptSigSize(_scriptPubKey: Script): bool =
+            false
+
+        override self.EstimateScriptSigSize(_scriptPubKey: Script): int =
+            raise <| NotSupportedException()
+
+        override self.CanCombineScriptSig(_scriptPubKey: Script, _a: Script, _b: Script): bool = 
+            false
+
+        override self.CombineScriptSig(_scriptPubKey: Script, _a: Script, _b: Script): Script =
+            raise <| NotSupportedException()
+
+        override self.IsCompatibleKey(pubKey: PubKey, scriptPubKey: Script): bool =
+            match CommitmentToLocalParameters.TryExtractParameters scriptPubKey with
+            | None -> false
+            | Some parameters -> parameters.LocalDelayedPubKey.RawPubKey() = pubKey
+
+
--- a/src/DotNetLightning.Core/Channel/CommitmentsModule.fs
+++ b/src/DotNetLightning.Core/Channel/CommitmentsModule.fs
@@ -1,5 +1,6 @@
 namespace DotNetLightning.Channel
 
+open System
 open NBitcoin
 
 open DotNetLightning.Utils
@@ -415,3 +416,83 @@ module internal Commitments =
                               OriginChannels = originChannels1 }
                 return [ WeAcceptedCommitmentSigned(nextMsg, nextCommitments) ;]
             }
+
+module ForceCloseFundsRecovery =
+    // The lightning spec specifies that commitment txs use version 2 bitcoin transactions.
+    [<Literal>]
+    let TxVersionNumberOfCommitmentTxs = 2u
+
+    let private check(thing: bool): Option<unit> =
+        if thing then
+            Some ()
+        else
+            None
+
+    let private tryGetObscuredCommitmentNumber (fundingOutPoint: OutPoint)
+                                               (transaction: Transaction)
+                                                   : Option<ObscuredCommitmentNumber> = option {
+        do! check (transaction.Version = TxVersionNumberOfCommitmentTxs)
+        let! txIn = Seq.tryExactlyOne transaction.Inputs
+        do! check (fundingOutPoint = txIn.PrevOut)
+        let! obscuredCommitmentNumber =
+            ObscuredCommitmentNumber.TryFromLockTimeAndSequence transaction.LockTime txIn.Sequence
+        return obscuredCommitmentNumber
+    }
+
+    let tryGetFundsFromLocalCommitmentTx (localParams: LocalParams)
+                                         (remoteParams: RemoteParams)
+                                         (fundingScriptCoin: ScriptCoin)
+                                         (localChannelPrivKeys: ChannelPrivKeys)
+                                         (network: Network)
+                                         (transaction: Transaction)
+                                             : Option<TransactionBuilder> = option {
+        let! obscuredCommitmentNumber =
+            tryGetObscuredCommitmentNumber
+                fundingScriptCoin.Outpoint
+                transaction
+        let localChannelPubKeys = localParams.ChannelPubKeys
+        let remoteChannelPubKeys = remoteParams.ChannelPubKeys
+        let commitmentNumber =
+            obscuredCommitmentNumber.Unobscure
+                localParams.IsFunder
+                localChannelPubKeys.PaymentBasepoint
+                remoteChannelPubKeys.PaymentBasepoint
+
+        let perCommitmentPoint =
+            localChannelPrivKeys.CommitmentSeed.DerivePerCommitmentPoint commitmentNumber
+        let localCommitmentPubKeys =
+            perCommitmentPoint.DeriveCommitmentPubKeys localChannelPubKeys
+        let remoteCommitmentPubKeys =
+            perCommitmentPoint.DeriveCommitmentPubKeys remoteChannelPubKeys
+
+        let transactionBuilder = network.CreateTransactionBuilder()
+
+        let toLocalScriptPubKey =
+            Scripts.toLocalDelayed
+                remoteCommitmentPubKeys.RevocationPubKey
+                localParams.ToSelfDelay
+                localCommitmentPubKeys.DelayedPaymentPubKey
+        let! toLocalIndex =
+            let toLocalWitScriptPubKey = toLocalScriptPubKey.WitHash.ScriptPubKey
+            Seq.tryFindIndex
+                (fun (txOut: TxOut) -> txOut.ScriptPubKey = toLocalWitScriptPubKey)
+                transaction.Outputs
+
+        let delayedPaymentPrivKey =
+            perCommitmentPoint.DeriveDelayedPaymentPrivKey
+                localChannelPrivKeys.DelayedPaymentBasepointSecret
+
+        transactionBuilder
+            .SetVersion(TxVersionNumberOfCommitmentTxs)
+            .Extensions.Add(CommitmentToLocalExtension())
+        return
+            transactionBuilder
+                .AddKeys(delayedPaymentPrivKey.RawKey())
+                .AddCoin(
+                    ScriptCoin(transaction, uint32 toLocalIndex, toLocalScriptPubKey),
+                    CoinOptions(
+                        Sequence = (Nullable <| Sequence(uint32 localParams.ToSelfDelay.Value))
+                    )
+                )
+    }
+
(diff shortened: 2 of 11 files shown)


src/DotNetLightning.Core/Channel/CommitmentsModule.fs
```diff
@@ -415,3 +416,80 @@ module internal Commitments =
                               OriginChannels = originChannels1 }
                 return [ WeAcceptedCommitmentSigned(nextMsg, nextCommitments) ;]
             }
+
+module ForceCloseFundsRecovery =
+    // The lightning spec specifies that commitment txs use version 2 bitcoin transactions.
+    let TxVersionNumberOfCommitmentTxs = 2u
```
joemphilips, line 422: You can specify `[<Literal>]` attribute for constants like this. Which will slightly improves performance.

src/DotNetLightning.Core/Utils/SeqConsumer.fs
```diff
@@ -0,0 +1,83 @@
...
+    let seqConsumer<'SeqElement> = SeqConsumerBuilder<'SeqElement>()
+
+    let NextInSeq<'SeqElement>(): SeqConsumer<'SeqElement, 'SeqElement> = {
+        Consume = fun (sequence: seq<'SeqElement>) ->
+            match Seq.tryHead sequence with
+            | None -> None
+            | Some value -> Some (Seq.tail sequence, value)
+    }
+
+    let AbortSeqConsumer<'SeqElement, 'T>(): SeqConsumer<'SeqElement, 'T> = {
```
joemphilips, line 65: Why not just name this `abort` ? And also required qualified access to `SeqConsumer` module?
I think it is more natural to access this function as `SeqConsumer.abort`. (It also helps to find the function with intellisense)

src/DotNetLightning.Core/Utils/SeqConsumer.fs
```diff
@@ -0,0 +1,83 @@
...
+                    seqConsumer.Consume sequence
+                with
+                | ex ->
+                    let subSeqConsumer = onException ex
+                    subSeqConsumer.Consume sequence
+        }
+
+
+module SeqConsumer =
+    let seqConsumer<'SeqElement> = SeqConsumerBuilder<'SeqElement>()
```
joemphilips, line 56: This should be defined in the `SeqConsumerCE` module. Or at lease like independent module as you do in `OptionCE`

src/DotNetLightning.Core/Utils/SeqConsumer.fs
```diff
@@ -0,0 +1,83 @@
...
+            match Seq.tryHead sequence with
+            | None -> None
+            | Some value -> Some (Seq.tail sequence, value)
+    }
+
+    let AbortSeqConsumer<'SeqElement, 'T>(): SeqConsumer<'SeqElement, 'T> = {
+        Consume = fun (_sequence: seq<'SeqElement>) -> None
+    }
+
+    type ConsumeAllError =
```
joemphilips, line 69: Define this in outside of the module. Also, why not just  name it`SeqConsumerError` ?

src/DotNetLightning.Core/Utils/SeqConsumer.fs
```diff
@@ -0,0 +1,83 @@
...
+        | SequenceEndedTooEarly
+        | SequenceNotReadToEnd
+
+    let ConsumeAll<'SeqElement, 'T>(sequence: seq<'SeqElement>) (seqConsumer: SeqConsumer<'SeqElement, 'T>)
```
joemphilips, line 73: Must be camelCase

src/DotNetLightning.Core/Utils/SeqConsumer.fs
```diff
@@ -0,0 +1,83 @@
+namespace DotNetLightning.Utils
+
+open ResultUtils
+open ResultUtils.Portability
+
+[<AutoOpen>]
+module SeqConsumerCE =
+    type SeqConsumer<'SeqElement, 'T> = {
```
joemphilips, line 8: An idea of `SeqConsumer` is quite neat and I like it. sounds like simple version of parser combinator.
Would you mind adding simple explanation document to the type?

src/DotNetLightning.Core/Channel/CommitmentToLocalExtension.fs
```diff
@@ -0,0 +1,131 @@
...
+
+open ResultUtils
+open ResultUtils.Portability
+
+type CommitmentToLocalParameters = {
+    ToSelfDelay: BlockHeightOffset16
+    LocalDelayedPubKey: DelayedPaymentPubKey
+}
+    with
+    static member TryExtractParameters (scriptPubKey: Script): Option<CommitmentToLocalParameters> =
```
joemphilips, line 18: This function needs a unit test. Or property test parameterized by two items to extract.

src/DotNetLightning.Core/Channel/CommitmentsModule.fs
```diff
@@ -415,3 +416,80 @@ module internal Commitments =
...
+
+    let check(thing: bool): Option<unit> =
```
joemphilips, line 424: Must be private?

src/DotNetLightning.Core/Channel/CommitmentsModule.fs
```diff
@@ -415,3 +416,80 @@ module internal Commitments =
...
+        if thing then
+            Some ()
+        else
+            None
+
+    let tryGetObscuredCommitmentNumber (fundingOutPoint: OutPoint)
```
joemphilips, line 430: must be private?

src/DotNetLightning.Core/Channel/CommitmentsModule.fs
```diff
@@ -415,3 +416,80 @@ module internal Commitments =
...
+                transaction.Outputs
+
+        let delayedPaymentPrivKey =
+            perCommitmentPoint.DeriveDelayedPaymentPrivKey
+                localChannelPrivKeys.DelayedPaymentBasepointSecret
+
+        transactionBuilder.SetVersion TxVersionNumberOfCommitmentTxs
+        |> ignore
+        transactionBuilder.Extensions.Add (CommitmentToLocalExtension())
+        transactionBuilder.AddKeys (delayedPaymentPrivKey.RawKey()) |> ignore
```
joemphilips, line 485: nit: I think you can make less verbose by chaining the methods instead of specifying `ignore` every time.

src/DotNetLightning.Core/Channel/CommitmentsModule.fs
```diff
@@ -415,3 +416,80 @@ module internal Commitments =
...
+                                           : Option<ObscuredCommitmentNumber> = option {
+        do! check (transaction.Version = TxVersionNumberOfCommitmentTxs)
+        let! txIn = Seq.tryExactlyOne transaction.Inputs
+        do! check (fundingOutPoint = txIn.PrevOut)
+        let! obscuredCommitmentNumber =
+            ObscuredCommitmentNumber.TryFromLockTimeAndSequence transaction.LockTime txIn.Sequence
+        return obscuredCommitmentNumber
+    }
+
+    let tryGetFundsFromLocalCommitmentTx (commitments: Commitments)
```
joemphilips, line 441: This needs a unit test.

joemphilips (changes requested):

src/DotNetLightning.Core/Channel/CommitmentToLocalExtension.fs
```diff
@@ -0,0 +1,131 @@
...
+        let ops =
+            scriptPubKey.ToOps()
+            // we have to collect it into a list and convert back to a seq
+            // because the IEnumerable that NBitcoin gives us is internally
+            // mutable.
+            |> List.ofSeq
+            |> Seq.ofList
```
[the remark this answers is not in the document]
canndrew: It doesn't work without that conversion back-and-forth. I think the problem was that `SeqConsumer` assumes that you can look at the head of a sequence without changing the sequence, but the `IEnumerable` returned by NBitcoin will pop the head when you do that. ie. this will fail:

```fsharp
let foo = Seq.head sequence
let bar = Seq.head sequence
assert (foo = bar)
```

src/DotNetLightning.Core/Utils/SeqConsumer.fs
canndrew: I named it `ConsumeAllError` because it's specific to the `ConsumeAll` method. We may want to have other kinds of errors related to `SeqConsumer`s (although there's none at the moment). For instance if we add a method which consumes a sequence but doesn't fail if the sequence isn't read to completion then that would have it's own error.

src/DotNetLightning.Core/Utils/SeqConsumer.fs
canndrew: That's what it is, yeah. Though I didn't think of that while I was writing it. Maybe I should rename it `SeqParser`?

knocte: can we also rename `'T` to something longer please?

src/DotNetLightning.Core/Utils/SeqConsumer.fs
```diff
@@ -0,0 +1,83 @@
...
+
+
+module SeqConsumer =
+    let seqConsumer<'SeqElement> = SeqConsumerBuilder<'SeqElement>()
+
+    let NextInSeq<'SeqElement>(): SeqConsumer<'SeqElement, 'SeqElement> = {
+        Consume = fun (sequence: seq<'SeqElement>) ->
+            match Seq.tryHead sequence with
+            | None -> None
+            | Some value -> Some (Seq.tail sequence, value)
```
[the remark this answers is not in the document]
canndrew: Fixed. Although I named the method `next`.

src/DotNetLightning.Core/Channel/CommitmentsModule.fs
canndrew: Fixed.

src/DotNetLightning.Core/Utils/SeqConsumer.fs
canndrew: Fixed. I agree that that looks a lot better.

src/DotNetLightning.Core/Utils/SeqConsumer.fs
canndrew: I moved it into `SeqConsumerCE`.

src/DotNetLightning.Core/Utils/SeqConsumer.fs
canndrew: Fixed.

src/DotNetLightning.Core/Channel/CommitmentToLocalExtension.fs
canndrew: I've added a unit test for this.

src/DotNetLightning.Core/Channel/CommitmentsModule.fs
canndrew: Fixed.

src/DotNetLightning.Core/Channel/CommitmentsModule.fs
canndrew: Fixed.

src/DotNetLightning.Core/Channel/CommitmentsModule.fs
canndrew: I couldn't see a (syntactically) simple way to do this without using lots of `|> fun transactionBuilder -> transationBuilder.method ...`, but that would be even more verbose. Maybe F# has some syntax for this that I'm not aware of (?)

src/DotNetLightning.Core/Channel/CommitmentsModule.fs
canndrew: I'm in the process of writing a test for this. However it involves settings up an awful lot of state (basically going through the whole process of creating a channel), so I haven't finished yet. I've also been learning how to use FsCheck and generators since the test requires so much arbitrary information.

src/DotNetLightning.Core/Channel/CommitmentsModule.fs
```diff
@@ -415,3 +416,80 @@ module internal Commitments =
...
+        transactionBuilder.Extensions.Add (CommitmentToLocalExtension())
+        transactionBuilder.AddKeys (delayedPaymentPrivKey.RawKey()) |> ignore
+        transactionBuilder.AddCoin(
+            ScriptCoin(transaction, uint32 toLocalIndex, toLocalScriptPubKey),
+            CoinOptions(
+                Sequence = (Nullable <| Sequence(uint32 commitments.LocalParams.ToSelfDelay.Value))
+            )
+        ) |> ignore
+
+        return transactionBuilder
```
joemphilips, line 494:
```suggestion
        return
            transactionBuilder
                .AddKeys(delayedPaymentPrivKey.RawKey())
                .AddCoin(
                    ScriptCoin(transaction, uint32 toLocalIndex, toLocalScriptPubKey), CoinOptions(
                        Sequence = (Nullable <| Sequence(uint32 commitments.LocalParams.ToSelfDelay.Value))
                    )
            )
```

src/DotNetLightning.Core/Channel/CommitmentsModule.fs
joemphilips: I've added a suggestion below. I'm not talking about something fancy, just regular syntax for both C# and F#.

src/DotNetLightning.Core/Utils/SeqConsumer.fs
joemphilips: > Maybe I should rename it SeqParser?

Yes, I think `SeqParser` is more easy to understand. consuming does not imply parsing.

tests/DotNetLightning.Core.Tests/CommitmentToLocalExtensionTests.fs
```diff
@@ -0,0 +1,38 @@
...
+open DotNetLightning.Channel
+
+open ResultUtils
+open ResultUtils.Portability
+
+[<Tests>]
+let commitmentToLocalExtensionTests =
+    testList "CommitmentToLocalExtensionTests" [
+        testCase "can extract parameters" <| fun _ ->
+            let rand = new Random()
```
joemphilips, line 17: nit: unnecessary `new`, `new` keyword in F# should be always coupled to an usage of `IDispoasble` , thus you should use `use` instead of `let` .So when you instantiate `Key` you should `use` it.

src/DotNetLightning.Core/Channel/CommitmentToLocalExtension.fs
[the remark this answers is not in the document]
joemphilips: hmm, OK.

src/DotNetLightning.Core/Channel/CommitmentsModule.fs
canndrew: *uugh..* Your code works, however my code didn't. The difference is that I had a space between `AddKeys` and the open-paren:

```
.AddKeys (delayedPaymentPrivKey.RawKey())
        ^ here
```

src/DotNetLightning.Core/Utils/SeqConsumer.fs
canndrew: Fixed. I've renamed it `SeqParser`. Also the generic argument is now called `'Value` instead `'T`.

src/DotNetLightning.Core/Channel/CommitmentsModule.fs
canndrew: Fixed.

tests/DotNetLightning.Core.Tests/CommitmentToLocalExtensionTests.fs
canndrew: Fixed.

canndrew:
I've written the test for `tryGetFundsFromLocalCommitmentTx`. Note that I also commented-out the test "check pre-computed transaction weights" because this test was (seemingly accidentally) not being run, and when it's made to run it fails.

src/DotNetLightning.Core/Utils/SeqParser.fs
```diff
@@ -0,0 +1,109 @@
+namespace DotNetLightning.Utils
+
+open ResultUtils
+open ResultUtils.Portability
+
+// A SeqParser is a parser for sequences. It wraps a function which takes a
```
joemphilips, line 6: nit: comments like this is better written with `///` rather than `//`. See https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/xml-documentation#comments-without-xml-tags

tests/DotNetLightning.Core.Tests/TransactionTests.fs
```diff
@@ -1,18 +1,155 @@
...
+            let extKey = new ExtKey()
+            NodeMasterPrivKey extKey
+        let remoteNodeSecret = remoteNodeMasterPrivKey.NodeSecret()
+        let remoteNodeId = remoteNodeSecret.NodeId()
+        let remoteChannelPrivKeys = remoteNodeMasterPrivKey.ChannelPrivKeys (rand.Next(1, 100))
+        let remoteChannelPubKeys = remoteChannelPrivKeys.ToChannelPubKeys()
+        let remoteDestPrivKey = new Key()
+        let remoteDestPubKey = remoteDestPrivKey.PubKey
+
+        let randomU256(): uint256 =
```
joemphilips, line 42: nit: you can use `NBitcoin.RandomUtils.GetUInt256`

canndrew:
I've fixed the nits and all the new warnings that my code was producing (which I hadn't noticed). I'll rebase this now to depend on the warnings-fixing PR.

canndrew:
This is now rebased off of the other PR.

knocte:
There are many unneeded commits here, can you squash some?

tests/DotNetLightning.Core.Tests/TransactionTests.fs
```diff
@@ -1,18 +1,146 @@
...
+        Expect.equal input.PrevOut.Hash (commitmentTx.GetHash()) "wrong prevout hash"
+        let expectedAmount =
+            let fullAmount = commitmentSpec.ToLocal.ToMoney()
+            let fee = commitmentTx.GetFee [| fundingScriptCoin |]
+            fullAmount - fee
+        let actualAmount =
+            commitmentTx.Outputs.[input.PrevOut.N].Value
+        Expect.equal actualAmount expectedAmount "wrong prevout amount"
+
+    (*
```
joemphilips, line 143: We should probably just remove this test.
It was originally ported from https://github.com/ACINQ/eclair/blob/f202587e9c71b2106194c0e6cabc8032f6f67df4/eclair-core/src/test/scala/fr/acinq/eclair/transactions/TransactionsSpec.scala#L111

But I guess it is better to add it again when we encounter the need.

joemphilips:
LGTM. Tell me you want me to merge.

src/DotNetLightning.Core/Channel/CommitmentsModule.fs
knocte: I was also surprised when I discovered this interesting behaviour from the F# compiler... for more info see this thread: https://github.com/fsprojects/fantomas/issues/1437

canndrew:
I've squashed all the fixup commits. This should be good to merge now.

src/ResultUtils/OptionCE.fs
```diff
@@ -0,0 +1,73 @@
+namespace ResultUtils
+
+open System
+
+[<AutoOpen>]
+module OptionCE =
+  type OptionBuilder() =
```
knocte, line 7: indentation is off here, should be 4 spaces, not 2

src/DotNetLightning.Core/Channel/CommitmentToLocalExtension.fs
```diff
@@ -0,0 +1,130 @@
...
+            false
+
+        override self.EstimateScriptSigSize(_scriptPubKey: Script): int =
+            raise <| System.NotSupportedException()
+
+        override self.CanCombineScriptSig(_scriptPubKey: Script, _a: Script, _b: Script): bool = 
+            false
+
+        override self.CombineScriptSig(_scriptPubKey: Script, _a: Script, _b: Script): Script =
+            raise <| System.NotSupportedException()
```
knocte, line 123: nit: no need for `System.` in these 3 lines above

canndrew: Fixed.

src/ResultUtils/OptionCE.fs
canndrew: Fixed.

knocte:
LGTM

PR #149, opened 2021-03-24
Fix build warnings

canndrew:
This PR fixes the build warnings.


canndrew:
Note that this conflicts with the other PR I have open (#148). Whichever gets merged first I'll just rebase the other.

canndrew:
I've added a commit to treat warnings as errors when building in release mode. This is in order to make CI fail so that we'll be prevented from letting new warnings make it onto master.

joemphilips:
Thanks for doing this. I just merge this PR first since it is easier to understand than #148

PR #150, opened 2021-03-26
Support retrieving funds from remotely force-closed channels

canndrew:
This PR adds a `tryGetFundsFromRemoteCommitmentTx` function, similar to `tryGetFundsFromLocalCommitmentTx` except that it recovers funds from a remote commitment tx. Currently, it only redeems the `to_remote` output of the commitment tx, ie. it doesn't implement revocation. However a future PR will expand this function to also redeem to `to_local` output when we have the remote's per-commitment-secret for the commitment tx.

This also expands the test for `tryGetFundsFromLocalCommitmentTx` to test both functions, since they re-use the same setup.

### DIFF
--- a/src/DotNetLightning.Core/Channel/CommitmentsModule.fs
+++ b/src/DotNetLightning.Core/Channel/CommitmentsModule.fs
@@ -439,6 +439,10 @@ module ForceCloseFundsRecovery =
         return obscuredCommitmentNumber
     }
 
+    /// This function returns a TransactionBuilder with the necessary inputs
+    /// added to the transaction to reclaim the funds from the supplied
+    /// commitment transaction. It is the caller's responsibility to add outputs
+    /// and fees to the TransactionBuilder before building the transaction.
     let tryGetFundsFromRemoteCommitmentTx (localParams: LocalParams)
                                           (remoteParams: RemoteParams)
                                           (fundingScriptCoin: ScriptCoin)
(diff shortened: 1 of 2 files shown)


joemphilips:
LGTM. I will merge tomorrow.

src/DotNetLightning.Core/Channel/CommitmentsModule.fs
```diff
@@ -439,6 +439,58 @@ module ForceCloseFundsRecovery =
         return obscuredCommitmentNumber
     }
 
+    let tryGetFundsFromRemoteCommitmentTx (localParams: LocalParams)
```
joemphilips, line 442: nit: It may be good to add an explanation that adding outputs to the returned `TransactionBuilder` is a caller's responsibility.
This also applies to the local counterpart (`tryGetFundsFromLocalCommitmentTx`)

tests/DotNetLightning.Core.Tests/TransactionTests.fs
```diff
@@ -140,6 +140,84 @@ let testList = testList "transaction tests" [
...
+            ForceCloseFundsRecovery.tryGetFundsFromRemoteCommitmentTx
+                remoteLocalParams
+                remoteRemoteParams
+                fundingScriptCoin
+                remoteRemotePerCommitmentSecrets
+                remoteRemoteCommit
+                remoteChannelPrivKeys
+                Network.RegTest
+                commitmentTx
+        let transactionBuilder = transactionBuilderOpt.Value
```
knocte, line 206: shouldn't this need a match? otherwise if it's None it could blow up here

joemphilips: This is a test so it does not really matter.
It is slightly better to assert by `Expect.isSome` before this line. But it is not a big deal.

canndrew: I added a couple of `Expect.isSome`s.

src/DotNetLightning.Core/Channel/CommitmentsModule.fs
canndrew: I added doc comments to the two functions.

knocte:
LGTM

Issue #153, opened 2021-04-11
CI issues


knocte:
I've seen a couple of issues lately with CI:
* Some lanes fail randomly, e.g.: the one for this recent commit: https://github.com/joemphilips/DotNetLightning/commit/4812785e7971b81cb763ba0d94a5de804854da73
* In our fork where we test several things, we have noticed that the push to nuget.org can happen even if the other CI lanes didn't pass, which is wrong.

joemphilips:
Let me investigate

joemphilips:
The CI failure itself is caused only in "windows && non-bouncycastle" build in Macaroons.
I remember I had similar issue once before.
I can just stop supporting Macaroons in case of non-bouncycastle build (Or maybe I should stop supporting NSec all together).

But before that I want to support .NET 5 and future versions, so please give me some time to research what is the best way to do so.

I'm pretty sure that this issue does not affect you as long as you are using bouncycastle build. So you can ignore it for now.

knocte:
> this issue does not affect you 

Yeah it doesn't matter for us technically, but it's noise when CI fails for reasons outside our control and the workaround is to just re-run CI.

> But before that I want to support .NET 5 and future versions, so please give me some time to research what is the best way to do so.

Sure, but DNL is a library, it doesn't have any executables. Therefore it can stay in .NETStandard form and .NET5 programs can use it just fine.

> In our fork where we test several things, we have noticed that the push to nuget.org can happen even if the other CI lanes didn't pass, which is wrong.

And how about this?

knocte:
> I can just stop supporting Macaroons in case of non-bouncycastle build

How about just run Macaroon tests in BC build? and don't include it in the nuget package for the nonBC build

joemphilips:
Hotfix merged in https://github.com/joemphilips/DotNetLightning/pull/155

In the future probably I must consider either
1. Stop supporting NSec build or
2. Fixed the test-failure issue on the windows.

PR #155, opened 2021-04-12
Stop publishing Macaroons in non-BouncyCastle build.

joemphilips:
* Since Macaroons test has annoying Bug which only occurs in "windows * non-bouncycastle build"
* See https://github.com/joemphilips/DotNetLightning/issues/153

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
         dotnet run --project tests/DotNetLightning.Core.Tests
     - name: Run other tests
       run: |
-        dotnet test
+        dotnet test --filter "FullyQualifiedName!~Macaroon" # see https://github.com/joemphilips/DotNetLightning/issues/153
 
   build_with_fsharp_from_mono:
     runs-on: ubuntu-20.04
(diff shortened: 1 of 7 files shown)


knocte:
> Stop publishing Macaroons in BouncyCastle build.

You mean in nonBC build? Because later you say `Since Macaroons test has an annoying bug which only occurs in "windows * non-bouncycastle" build`

joemphilips:
> You mean in nonBC build? Because later you say Since Macaroons test has an annoying bug which only occurs in "windows * non-bouncycastle" build

You're right, fixed.

knocte:
> You're right, fixed.

But you didn't fix the commit message 😅

joemphilips:
> But you didn't fix the commit message

whoops. now fixed. I also squashed commits into one.

knocte:
> we renamed this define to Portability recently

you didn't fix this?

joemphilips:
> you didn't fix this?

I'm not sure what you meant.  [We still use `BouncyCastle` as a compile-time variable name](https://github.com/joemphilips/DotNetLightning/blob/179788a33489746fef7bb3de53050514b74f567f/Directory.Build.props#L18)
And that is not affected by this change.

knocte:
Oops you're right, I didn't remember the more specific details of https://github.com/joemphilips/DotNetLightning/commit/3d06651bda42676b48c0a7cbff3aa656e950593f

PR #158, opened 2021-05-19
Fix key-tweaking bug in BouncySecp256k1

canndrew:
The function `BouncyCastleSecp256k1.tweakKey` would return a bogus result if the tweaked key contained leading zero bytes. This is because it would first convert the result from a `BigInteger` to an array, truncating any leading zero bytes, then write the array to the target `Span<byte>`, ignoring the possibility that the array is too small due to truncated bytes.

This PR fixes the issue by writing the resulting array to the correct offset into the span and zeroing the head of the span.


canndrew:
Key tweaking is used when deriving the per-commitment keys from the basepoint keys. "tweaking" is the terminology used in the lightning spec, `NBitcoin` and in the rust `secp256k1` library, so I don't think it makes sense to change the name.

joemphilips (approved):

PR #160, opened 2021-05-20
Make two functions public

aarani:
Move/publicize ChannelError.feeRateMismatch and ForceCloseFundsRecovery.tryGetObscuredCommitmentNumber

These two functions are useful outside of DotNetLightning.

### DIFF
--- a/src/DotNetLightning.Core/Channel/CommitmentsModule.fs
+++ b/src/DotNetLightning.Core/Channel/CommitmentsModule.fs
@@ -449,9 +449,9 @@ module ForceCloseFundsRecovery =
                     (lockTime.ToString())
                     (sequence.ToString())
 
-    let private tryGetObscuredCommitmentNumber (fundingOutPoint: OutPoint)
-                                               (transaction: Transaction)
-                                                   : Result<ObscuredCommitmentNumber, ValidateCommitmentTxError> = result {
+    let tryGetObscuredCommitmentNumber (fundingOutPoint: OutPoint)
+                                       (transaction: Transaction)
+                                           : Result<ObscuredCommitmentNumber, ValidateCommitmentTxError> = result {
         if transaction.Version <> TxVersionNumberOfCommitmentTxs then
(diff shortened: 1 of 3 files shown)


joemphilips (approved):

src/DotNetLightning.Core/Utils/Primitives.fs
```diff
@@ -338,12 +338,18 @@ module Primitives =
         member this.AsNBitcoinFeeRate() =
             this.Value |> uint64 |> (*)4UL |> Money.Satoshis |> FeeRate
 
+        member this.MismatchRatio (other: FeeRatePerKw) =
```
joemphilips, line 341: Would you mind adding a doc-comment with `///` for this method that the returned value is agnostic to the absolute feerate and it is always positive? It may be good to add a reference to the wiki (https://en.wikipedia.org/wiki/Relative_change_and_difference) too.

knocte:
@joemphilips ready for review again.

src/DotNetLightning.Core/Channel/CommitmentsModule.fs
```diff
@@ -449,9 +449,9 @@ module ForceCloseFundsRecovery =
                     (lockTime.ToString())
                     (sequence.ToString())
 
-    let private tryGetObscuredCommitmentNumber (fundingOutPoint: OutPoint)
-                                               (transaction: Transaction)
-                                                   : Result<ObscuredCommitmentNumber, ValidateCommitmentTxError> = result {
+    let tryGetObscuredCommitmentNumber (fundingOutPoint: OutPoint)
```
joemphilips, line 452: Why it has to be public? The reason is not written in commit msg.

canndrew: It's used in geewallet to identify commitment transactions that have been broadcast to the blockchain by the remote peer.

joemphilips: Fair enough.

PR #161, opened 2021-05-26
Remove pre-channel states

canndrew:
This PR removes the pre-channel-establishment states from `ChannelState`, factoring them into their own types.

In the current design of DNL there is a type `Channel` which contains an interior `ChannelState`, and we perform operations on channels by calling `executeCommand` with a `ChannelCommand`. Some of these states only represent channels prior to channel establishment.  These states only support a single `ChannelCommand` and it's nonsense to either use a different command or to use that command when the channel is in a different state.

For instance, if we have a channel in the `WaitForAcceptChannel` state then the only command we can perform is `ApplyAcceptChannel` when we receive the corresponding `accept_channel` message. It doesn't make sense to attempt to (for example) close a channel in the `WaitForAcceptChannel` state because the channel hasn't been opened yet (ie. funding tx signatures haven't been exchanged yet). Nor does it make sense to attempt to perfom `ApplyAcceptChannel` on a channel which is any state other than `WaitForAcceptChannel` since an `accept_channel` message is only received/applied at a certain point in the channel setup routine.

Channels in these states also don't need to be saved to a wallet since there are no funds at stake and the lightning spec requires that these channels are forgotten in the case of an error or a breakdown in connectivity.

As such, we can simplify the API for the consumer of DNL, and make the code more robustly-typed, by representing these pre-channel states as their own types which each support a single operation to apply a message and advance to the next state. This makes it impossible to (for instance) attempt to apply an `accept_channel` message to channel that is already operational. On the consumer side (specifically in geewallet) this removes the need for a lot of asserts and unwraps which dynamically check that a channel is in the expected state, has a channel id, has commitments and a funding tx, etc. since these conditions can now be enforced by types.

### DIFF
--- a/src/DotNetLightning.Core/Channel/Channel.fs
+++ b/src/DotNetLightning.Core/Channel/Channel.fs
@@ -20,6 +20,7 @@ type ChannelWaitingForFundingSigned = {
     RemoteNodeId: NodeId
     NodeSecret: NodeSecret
     Network: Network
+    FundingTxMinimumDepth: BlockHeightOffset32
     WaitForFundingSignedData: WaitForFundingSignedData
 } with
     member self.ApplyFundingSigned (msg: FundingSignedMsg)
@@ -73,6 +74,7 @@ type ChannelWaitingForFundingSigned = {
             NodeSecret = self.NodeSecret
             State = WaitForFundingConfirmed nextState
             Network = self.Network
+            FundingTxMinimumDepth = self.FundingTxMinimumDepth
         }
         return state.FundingTx, channel
     }
@@ -84,6 +86,7 @@ and ChannelWaitingForFundingCreated = {
     RemoteNodeId: NodeId
     NodeSecret: NodeSecret
     Network: Network
+    FundingTxMinimumDepth: BlockHeightOffset32
     WaitForFundingCreatedData: WaitForFundingCreatedData
 } with
     member self.ApplyFundingCreated (msg: FundingCreatedMsg)
@@ -155,6 +158,7 @@ and ChannelWaitingForFundingCreated = {
             NodeSecret = self.NodeSecret
             Network = self.Network
             State = WaitForFundingConfirmed nextState
+            FundingTxMinimumDepth = self.FundingTxMinimumDepth
         }
         return msgToSend, channel
     }
@@ -222,6 +226,7 @@ and ChannelWaitingForFundingTx = {
             RemoteNodeId = self.RemoteNodeId
             NodeSecret = self.NodeSecret
             Network = self.Network
+            FundingTxMinimumDepth = state.LastReceived.MinimumDepth
             WaitForFundingSignedData = waitForFundingSignedData
         }
         return nextMsg, channelWaitingForFundingSigned
@@ -272,6 +277,7 @@ and Channel = {
     NodeSecret: NodeSecret
     State: ChannelState
     Network: Network
+    FundingTxMinimumDepth: BlockHeightOffset32
  }
         with
         static member NewOutbound(config: ChannelConfig,
@@ -330,6 +336,7 @@ and Channel = {
                                   network: Network,
                                   remoteNodeId: NodeId,
                                   inputInitFundee: InputInitFundee,
+                                  minimumDepth: BlockHeightOffset32,
                                   openChannelMsg: OpenChannelMsg
                                  ): Result<AcceptChannelMsg * ChannelWaitingForFundingCreated, ChannelError> =
             result {
@@ -343,7 +350,7 @@ and Channel = {
                     MaxHTLCValueInFlightMsat = localParams.MaxHTLCValueInFlightMSat
                     ChannelReserveSatoshis = localParams.ChannelReserveSatoshis
                     HTLCMinimumMSat = localParams.HTLCMinimumMSat
-                    MinimumDepth = config.ChannelHandshakeConfig.MinimumDepth
+                    MinimumDepth = minimumDepth
                     ToSelfDelay = localParams.ToSelfDelay
                     MaxAcceptedHTLCs = localParams.MaxAcceptedHTLCs
                     FundingPubKey = channelKeys.FundingPrivKey.FundingPubKey()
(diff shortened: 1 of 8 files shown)


knocte:
@joemphilips can you review/merge PR 162 before this one please?

joemphilips:
@canndrew Would you point me to the code using this API? I want to understand how this improves the type safety.

canndrew:
With these changes you can't attempt to:
  * Get the channel id of a channel that doesn't have a channel id yet.
  * Get the funding tx of a channel that doesn't have a funding tx yet.
  * Get the commitments of a channel that doesn't have commitments yet.
  * Apply an accept channel message to a channel that has already received an accept channel message.
  * Apply an accept channel message to an incoming channel.
  * Apply a funding created message to a channel that has already received a funding created message.
  * Apply a funding created message to an outgoing channel.
  * Apply a funding signed message to a channel that has already received a funding signed message.
  * Apply a funding signed message to an incoming channel.
  * etc.

This is all enforced at compile-time by the type-system. So if there's any way that the DNL-consumer's code could execute which would attempt to do one of things then the code won't compile and they'll be forced the re-write their code so that it's guaranteed to perform the channel-setup routine correctly (or at least more-correctly than they might have otherwise).

The changes I've made to geewallet for this can be found by diffing these branches:

  0. [ln-m7-before-dnl-channel-refactor](https://gitlab.com/canndrew/geewallet/-/commits/ln-m7-before-dnl-channel-refactor)
  1. [ln-m7-with-dnl-channel-refactor](https://gitlab.com/canndrew/geewallet/-/commits/ln-m7-with-dnl-channel-refactor)

For example, there's [this commit](https://gitlab.com/canndrew/geewallet/-/commit/3ff6a22b3f5ffaa2bb88fc3afd4216fa44d13126).  As you can see, this removes a lot of `UnwrapOption` and `failwith` calls and replaces a lot of uses of `Option<Something>` with just `Something`. We don't have to assert that things are in the state that we expect them to be when we can just type-check that they are.

knocte:
@joemphilips what do you think?

joemphilips:
Well, the way `Channel` module used here has become quite different from what I first expected.
My first intention is that users don't have to know the detail of the state (and events or any other inner mechanism) of the channel.

My version already fails to achieve this goal because I didn't think carefully about how to perform side-effects, and ended up that user must check each events and run side effects by their own.
Probably I should have prepared something like `Cmd` in elmish.

So this PR surely improves the type safetyness, I think it does not improve in terms of abstracting and hiding the inner Channel workings from users.

If you have a plan to create some abstraction layer on top of this API and if you could include that in DNL side, then I'm happy to merge this PR. otherwise, well ... I don't know, I still can merge it but the Channel module becomes quite useless as a reusable "library" anymore.

knocte:
> If you have a plan to create some abstraction layer on top of this API and if you could include that in DNL side

Maybe this is as easy as moving some of our geewallet code into DNL?

canndrew:
@joemphilips:

This PR is the first of a larger in-progress refactor to make DNL easier to use on the consumer side and fix some bugs along the way.

> My version already fails to achieve this goal because I didn't think carefully about how to perform side-effects, and ended up that user must check each events and run side effects by their own.

My plan here is to replace `executeCommand`/`applyEvent` with state-transition functions and remove `ChanneCommand`/`ChannelEvent` entirely. For instance, the process now for sending an HTLC is to:

 a) call `executeCommand` with `ChannelCommand.AddHTLC`
 b) search the returned `List<ChannelEvent>` and find the `ChannelEvent.WeAcceptedOperationAddHTLC` to get the message we need to send (or `failwith` if we don't find it).
 c) fold over the returned `List<ChannelEvent>` and call `applyEvent` with each event to get the new channel object.

This can be replaced with a method on `Channel`:

```fsharp
type Channel =
    member self.AddHTLC (htlc: OperationAddHTLC): Channel * Result<UpdateAddHTLCMsg, ChannelError> =
        ...
```

This is much simpler/safer from the consumer's point-of-view since it's just a single method call which returns the new channel object and is guaranteed to either return the message or an error. It's also simpler internally to DNL since the `executeCommand` and `applyEvent` logic for `update_add_htlc` is concatenated into the `AddHTLC` method rather than having it split between two places.

Another thing we need to do on the geewallet side is to serialize the channel's internal state to the wallet. However not all the state should be serialized. For instance behavioural runtime options like `MaxFeeRateMismatchRatio` or `MaxClosingNegotiationIterations` aren't really part of the state and can be passed in when we reload/reconnect the channel from a file. More importantly, if update messages have been exchanged but not yet signed when the TCP stream disconnects these messages should be forgotten when we reconnect. So part of my refactor plan is to identify the core, cryptographically-signed component of the channel state and seperate it out from the transient runtime state so that it can be saved/loaded.

> So this PR surely improves the type safetyness, I think it does not improve in terms of abstracting and hiding the inner Channel workings from users.

As long as the user has to handle the messaging side of things themselves then it's useful to expose some aspects of the channel's state. For instance, since we're the ones passing-in/extracting the channel-setup messages to/from DNL then it's useful for us to know that the channel is in the channel-setup state that we expect. It doesn't make it any easier to hide this away from us since we just have to replicate this state information on the geewallet side anyway.

There could be a higer-level APIs which hide this though. I imagine a next-higher-level API looking something like this (for the funder side):

```fsharp
// An encrypted network connection to a remote peer.
type ConnectedPeer =
    // Start opening a channel. Exchanges open/accept channel messages with the
    // peer and returns a `PendingOutgoingChannel` which can be supplied with a
    // funding transaction to continue construction of the channel.
    member self.OpenChannel
                (params: OpenChannelParams)
                : Async<Result<PendingOutgoingChannel, OpenChannelError>>

// A channel which we have started opening, but which we have yet to supply
// with a funding transaction.
type PendingOutgoingChannel =
    // The value of the funding output we must provide.
    member self.FundingAmount(): Money

    // The destination of the funding transaction we must provide.
    member self.FundingDestination(): IDestination

    // Provide the funding tx and continue opening the channel. Exchanges
    // funding created/signed messages with the peer then broadcasts the
    // transaction.
    member self.FundChannel
                (fundingTx: Transaction)
                : Async<Result<Channel, FundChannelError>>
```

In fact, this is pretty close to what we have in geewallet at the moment. Since the user of this API would no longer be concerned with handling individual messages, they don't need to be concerned with performing the individual state-transitions either.

joemphilips:
Yeah, OK I understand your point. I just wanted to let you know my overall intention of the `Channel` module.

The higher-API you have mentioned makes sense to me. I'm expecting you to make another PR in the future to incorporate those part in the DNL, but there is no hurry.

joemphilips:
LGTM. I will just merge now.

PR #162, opened 2021-05-27
Core/Channel: function to calc outpoints of commitTx

aarani:
Currently, we need to use a couple of internal functions to
calculate commitment outpoints amounts so we want to have this
to not only make a public API for this but also simplify it and
reduce the amount of duplicated code

We use this to calculate the exact amount of toRemote and toLocal
output to then be able to calculate the watch tower reward based
on that.

### DIFF
--- a/src/DotNetLightning.Core/Channel/Commitments.fs
+++ b/src/DotNetLightning.Core/Channel/Commitments.fs
@@ -10,6 +10,7 @@ open DotNetLightning.Serialization.Msgs
 
 open ResultUtils
 open ResultUtils.Portability
+open DotNetLightning.Transactions
 
 type LocalChanges = {
     Proposed: IUpdateMsg list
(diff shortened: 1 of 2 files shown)


src/DotNetLightning.Core/Channel/Commitments.fs
```diff
@@ -186,3 +193,46 @@ type Commitments = {
             match remoteSigned, localSigned with
             | Some _, Some htlcIn -> htlcIn.Add |> Some
             | _ -> None
+
+        static member RemoteCommitAmount (remoteParams: RemoteParams)
```
joemphilips, line 197: I think these methods does not have to be static.

src/DotNetLightning.Core/Channel/Commitments.fs
```diff
@@ -186,3 +193,46 @@ type Commitments = {
...
+                     |> Money.Satoshis) - commitFee
+                else
+                    (remoteCommit.Spec.ToLocal.Satoshi
+                     |> Money.Satoshis) - commitFee,
+                    (remoteCommit.Spec.ToRemote.Satoshi
+                     |> Money.Satoshis)
+
+            {Amounts.ToLocal = toLocalAmount; ToRemote = toRemoteAmount}
+
+        static member LocalCommitAmount (localParams: LocalParams)
```
joemphilips, line 219: This too.

PR #163, opened 2021-05-28
Fix "build_with_fsharp_from_mono" CI lane

knocte:
Somehow, the "ResolveReferences" target was not found anymore by
the msbuild tool used in this lane, but we don't need this target
when using Mono because this workaround is only used when using
"dotnet" to create the nuget package.


knocte:
@joemphilips FYI this fixes the build issue that happened in PR 162

PR #164, opened 2021-06-07
Update message formats for open/accept channel

canndrew:
For more recent versions of the lightning spec open/accept channel messages now use a tlv stream at the end of the message to indicate the presence of an upfront shutdown script. This turns out to be backwards-compatible, in that the on-the-wire format turns out to be identical.

If `option_upfront_shutdown_script` was advertised by both parties, then the shutdown script field/TLV is mandatory (but may be length `0` to indicate no shutdown script), otherwise the field is optional (but still allowed to be present). As such, it seems simpler and more bug-proof to always include the field and to set it to a length-`0` script if we don't have a shutdown script. Hence the changes to the test.

### DIFF
--- a/src/DotNetLightning.Core/Serialization/Msgs/Msgs.fs
+++ b/src/DotNetLightning.Core/Serialization/Msgs/Msgs.fs
@@ -513,7 +513,7 @@ type OpenChannelMsg = {
     mutable HTLCBasepoint: HtlcBasepoint
     mutable FirstPerCommitmentPoint: PerCommitmentPoint
     mutable ChannelFlags: uint8
-    mutable ShutdownScriptPubKey: OptionalField<Script>
+    mutable TLVs: array<OpenChannelTLV>
 }
 with
     interface IChannelMsg
@@ -537,9 +537,10 @@ with
             this.HTLCBasepoint <- ls.ReadHtlcBasepoint()
             this.FirstPerCommitmentPoint <- ls.ReadPerCommitmentPoint()
             this.ChannelFlags <- ls.ReadUInt8()
-            this.ShutdownScriptPubKey <-
-                if (ls.Position = ls.Length) then None else
-                ls.ReadWithLen() |> Script |> Some
+            this.TLVs <-
+                ls.ReadTLVStream()
+                |> Array.map OpenChannelTLV.FromGenericTLV
+
         member this.Serialize(ls) =
             ls.Write(this.Chainhash, true)
             ls.Write(this.TemporaryChannelId.Value, true)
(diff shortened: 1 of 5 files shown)


knocte:
@canndrew why is the commit message not the same as the PR description? The PR description looks to have more info so let's use that in the commit message too.

canndrew:
@knocte: Fixed. There's also a follow-up PR here: https://github.com/joemphilips/DotNetLightning/pull/165

src/DotNetLightning.Core/Serialization/Msgs/Msgs.fs
```diff
@@ -597,9 +608,18 @@ with
...
             this.ShutdownScriptPubKey <-
-                if (ls.Position = ls.Length) then None else
-                ls.ReadWithLen() |> Script |> Some
+                Seq.choose
+                    (function
+                        | AcceptChannelTLV.UpfrontShutdownScript script -> Some script
+                        | _ -> None
+                    )
+                    tlvs
+                |> Seq.tryExactlyOne
```
joemphilips, line 621: I don't think discarding `Unknown` fields like this is good idea.
The fact that "There is an Unknown TLV field sent from other party" is itself valuable and application might want to tell the user about it. Or it may want to pass the data to another application as-is.

I suggest to have a field `TLVs: AcceptChannelTLV[]` in `AcceptChannelMsg`  and add new property
```fs
this.ShutdownScriptPubKey: Script option =
    this.TLVs |> Seq.choose  // ...
```

knocte: @joemphilips updated

tests/DotNetLightning.Core.Tests/Generators/Msgs.fs
```diff
@@ -152,7 +155,10 @@ let acceptChannelGen =
         <*> delayedPaymentBasepointGen
         <*> htlcBasepointGen
         <*> perCommitmentPointGen
-        <*> (Gen.optionOf pushScriptGen)
+        <*> gen {
+                let! shutdownScript = (Gen.optionOf pushScriptGen)
+                return [| AcceptChannelTLV.UpfrontShutdownScript shutdownScript |]
```
joemphilips, line 160: TLV fields generator should include GenericTLV field. 
So this generator should be something like
```fs
Gen.oneOf [
    genericTLVGen([ 0UL ])
    gen {
      let! shutdownScript = (Gen.optionOf pushScriptGen)
      return [| AcceptChannelTLV.UpfrontShutdownScript shutdownScript |]
  }
]
```

knocte: @joemphilips updated

joemphilips:
LGTM

PR #165, opened 2021-06-07
Refactor shutdown scripts

canndrew:
This builds off of the previous PR here: https://github.com/joemphilips/DotNetLightning/pull/164

Previously the local shutdown script was stored in two places in the channel and also provided as an argument when initiating a shutdown.  This made it possible to send invalid shutdown messages if the shutdown scripts provided by the user at different times were different. Also, shutdown scripts have to be in certain forms, but the only place that they were validated was when initiating a shutdown.

There's now a `ShutdownScriptPubKey` type which wraps a `Script` and enforces that the script is a valid shutdown script. This is used throughout the code for all scripts which are shutdown scripts, which forces us to check their validity at all required points.

The `DefaultFinalScriptPubKey` field has been removed from `LocalParams` so that we only store the local shutdown script in (at most) one place in the channel data structure. This shutdown script is an `Option`, and can be `None` in the case that we didn't specify a shutdown script to give to the remote peer when creating the channel. As such as we still pass a shutdown script as an argument when closing a channel but we also check that it matches the recorded shutdown script that we previously gave the remote peer in the case that it's `Some`.

### DIFF
--- a/src/DotNetLightning.Core/Serialization/LightningStream.fs
+++ b/src/DotNetLightning.Core/Serialization/LightningStream.fs
@@ -5,6 +5,9 @@ open System.IO
 open DotNetLightning.Utils
 open NBitcoin
 
+open ResultUtils
+open ResultUtils.Portability
+
 type Scope(openAction: Action, closeAction: Action) =
     let _close = closeAction
     do openAction.Invoke()
@@ -416,4 +419,9 @@ type LightningReaderStream(inner: Stream) =
             rest <- int32 this.Length - int32 this.Position
         result |> Seq.toArray
          
-        
+    member this.ReadShutdownScriptPubKey(): ShutdownScriptPubKey =
+        let script = this.ReadScript()
+        match ShutdownScriptPubKey.TryFromScript script with
+        | Ok shutdownScript -> shutdownScript
+        | Error errorMsg ->
+            raise <| FormatException(errorMsg)
--- a/src/DotNetLightning.Core/Utils/Primitives.fs
+++ b/src/DotNetLightning.Core/Utils/Primitives.fs
@@ -452,3 +452,41 @@ module Primitives =
         | SortedPlain = 0uy
         | ZLib = 1uy
 
+    type ShutdownScriptPubKey = private {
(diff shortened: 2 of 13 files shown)


src/DotNetLightning.Core/Utils/Primitives.fs
```diff
@@ -452,3 +452,41 @@ module Primitives =
         | SortedPlain = 0uy
         | ZLib = 1uy
 
+    type ShutdownScriptPubKey = private {
+        Thingo: Script
```
knocte, line 456: Thingo?

canndrew: Oops. I had a commit to fix that but I forgot to include/squash it.

canndrew: Fixed.

knocte: is it aussie slang? :D

knocte:
@joemphilips CI is green after your merge.

src/DotNetLightning.Core/Serialization/LightningStream.fs
```diff
@@ -416,4 +419,9 @@ type LightningReaderStream(inner: Stream) =
             rest <- int32 this.Length - int32 this.Position
         result |> Seq.toArray
          
-        
+    member this.ReadShutdownScriptPubKey(): ShutdownScriptPubKey =
+        let script = this.ReadScript()
+        match ShutdownScriptPubKey.TryFromScript script with
+        | Ok shutdownScript -> shutdownScript
+        | Error errorMsg -> failwith errorMsg
```
joemphilips, line 426: Should raise `FormatException`

joemphilips (changes requested):

src/DotNetLightning.Core/Serialization/LightningStream.fs
canndrew: Good catch.

Fixed. And I've rebased the PR.

joemphilips:
LGTM

PR #166, opened 2021-06-07
Tx builder improvements

knocte:
* Core: be 100% sure to not have randomness in TransactionBuilder
* Core/Channel: use deterministic TransactionBuilder
* Use createDeterministicTransactionBuilder in makeCommitTx

### DIFF
--- a/src/DotNetLightning.Core/Transactions/Transactions.fs
+++ b/src/DotNetLightning.Core/Transactions/Transactions.fs
@@ -249,6 +249,10 @@ module Transactions =
 
     [<AutoOpen>]
     module Constants =
+        // The lightning spec specifies that commitment txs use version 2 bitcoin transactions.
+        [<Literal>]
+        let TxVersionNumberOfCommitmentTxs = 2u
+
         // ------- From eclair ---------
         [<Literal>]
         let COMMITMENT_TX_BASE_WEIGHT = 724UL
@@ -349,7 +353,7 @@ module Transactions =
                 remotePaymentBasepoint
 
     /// Sort by BOLT 3: Compliant way (i.e. BIP69 + CLTV order)
-    let sortTxOut (txOutsWithMeta: (TxOut * _) list) =
+    let sortTxOut (txOutsWithMeta: (TxOut * Option<UpdateAddHTLCMsg>) list) =
         txOutsWithMeta
         |> List.sortBy(fun txom -> { SortableTxOut.TxOut = (fst txom);UpdateAddHTLC = (snd txom) })
         |> List.map(fst)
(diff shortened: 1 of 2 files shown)


knocte:
@joemphilips this is ready for review now

src/DotNetLightning.Core/Transactions/Transactions.fs
```diff
@@ -348,7 +353,7 @@ module Transactions =
                 remotePaymentBasepoint
 
     /// Sort by BOLT 3: Compliant way (i.e. BIP69 + CLTV order)
-    let sortTxOut (txOutsWithMeta: (TxOut * _) list) =
+    let sortTxOut (txOutsWithMeta: List<TxOut * Option<UpdateAddHTLCMsg>>) =
```
joemphilips, line 356: nit: I prefer `'T list` style to avoid confusion with `ResizeArray`

src/DotNetLightning.Core/Transactions/Transactions.fs
```diff
@@ -403,18 +408,31 @@ module Transactions =
...
+                List.fold
+                    (fun ((txb, outAmount): TransactionBuilder * Money) (txOut: TxOut) ->
+                        txb.Send(txOut.ScriptPubKey, txOut.Value) |> ignore
+                        (txb, outAmount + txOut.Value)
+                    )
+                    (txb, Money(0UL))
+                    (sortTxOut txOuts)
+            let actualFee = inputInfo.Amount - outAmount
+            txb.SendFees actualFee |> ignore
+            txb.SetLockTime lockTime |> ignore
```
joemphilips, line 434: nit: chaining methods are more concise than `ignore` ing every time.
e.g.
```fs
txb
    .SendFees(actualFee)
    .SetLocktime(lockTime)
    |> ignore
```
Afterall that is what the fluent API is designed for.

joemphilips:
LGTM besides several nits.

joemphilips (dismissed):

src/DotNetLightning.Core/Transactions/Transactions.fs
knocte: we'll change it, but I'm curious why would `List<'T>` be confused with ResizeArray?

knocte:
Fixed.

src/DotNetLightning.Core/Transactions/Transactions.fs
joemphilips: Because `ResizeArray<T>` is just an alias for `System.Collections.Generic.List<T>`.
`List<T>` is ambiguous, we can not tell if it is `FSharp.Colletions.List<T>` or `System.Collections.Generic.List<T>`
`T list` is always `FSharp.Collections.List<T>`

knocte: Not sure I agree with that take (I'll give my rationale later), but why `'T list` instead of `list<'T>`?

joemphilips: > but why 'T list instead of list<'T>?

No big reason for this, just because it feels natural to post-fix the type parameter with the wrapper type in case the wrapper type is lowercase.
i. e.
`'T option`, `Option<'T>` feels more natural than `option<'T>` , `'T Option`

but again, it is not a big deal.

knocte: Ok, this is my take (which you might not agree with):

- I generally prefer APIs to follow .NET design guidelines, which means PascalCase for methods, type names, etc. Using lowercase for type names or methods, like some F#-ers do, strikes to me as very unwelcoming to C# people. I know in this case the distinction between list and List is useful because one guarantees immutability, but IMO the API consistency with C# is more important (and you can still enforce immutability via not allowing to open System.Collections.Generic at review time).
- Writing sometimes `'Foo Bar` and other times `Bar<'Foo>` is inconsistent to me. Most F# documentation advocates for the latter format so I always try to use the latter, which also looks less ambiguous and confuses C# devs less.

(My motivation here is mainly the thought that the group of C# devs that might join us into the F# club can be much bigger than the group of FP devs that might come from other langs. This is also why I don't use paket myself, but just nuget, etc.)

PR #167, opened 2021-06-08
Remove unused types from ChannelTypes module

canndrew:
The types `WaitForOpenChannelData`, `WaitForFundingInternalData`, and `WaitForRemotePublishFutureCommitmentData` were defined but not used at all. They have been removed.


joemphilips:
LGTM

PR #168, opened 2021-06-09
Core: return peer's closing_signed msg if accepted

aarani:
According to lightning spec, if the receiver agrees with
the fee, it should reply with a closing_signed with the
same fee_satoshis value so to transport this message we
need to pass it inside MutualClosePerformed event,
to be transported to the other peer.

Source:
https://github.com/lightningnetwork/lightning-rfc/blob/master/02-peer-protocol.md#requirements-6


joemphilips:
LGTM

PR #182, opened 2022-04-05
Core: FeatureBits.SetFeature should not mutate

aarani:
FeatureBits.SetFeature should return a new copy
when called to prevent bugs like
https://github.com/nblockchain/DotNetLightning.Kiss/commit/7caba22f29e5f3f2465f677a23ba1c063a822d10.


joemphilips (dismissed):
LGTM

src/DotNetLightning.Core/Serialization/Features.fs
```diff
@@ -178,13 +178,13 @@ module internal Feature =
         
         
 [<StructuredFormatDisplay("{PrettyPrint}")>]
-type FeatureBits private (bitArray) =
-    member val BitArray: BitArray = bitArray with get, set
+type FeatureBits private (bitArray: BitArray) =
+    member __.BitArray
```
knocte, line 182: @aarani can this be made internal?

src/DotNetLightning.Core/Serialization/Features.fs
```diff
@@ -178,13 +178,13 @@ module internal Feature =
...
+        with get() =
```
knocte, line 183: @aarani please note, I was actually wrong when I told you in telegram that `val` implies mutable state by default; but anyway, let's leave this like it is now, as it makes it consistent with the ByteArray property

knocte:
PR title and description should change because as it is now, it looks like if this PR was a refactoring. However, even if 1 of the commits in the PR is a refactoring (the 2nd one), the 1st one changes behaviour, therefore it cannot be considered a refactoring anymore. Let's edit PR details to align with the 1st commit instead.

Thanks

knocte:
LGTM, will merge later.

joemphilips (approved):

Issue #179, opened 2022-02-01
Replace Aether.fs with FSharpPlus (or deprecate)?


knocte:
See https://fsprojects.github.io/FSharpPlus/lens.html

joemphilips:
I'm not against it, `Aether.fs` style is kind of tedious.

In the hindsight I think lens/prism was unnecessary abstraction, so I prefer rather dropping it altogether.

PR #183, opened 2022-04-05
Core: remove usage of Aether helper (Lens/Prism)

aarani:
Fixes https://github.com/joemphilips/DotNetLightning/issues/179


joemphilips (approved):
LGTM (ideally PeerChannelEncryptor should care more about performance, but this change is definitly better than before)

PR #184, opened 2022-04-13
Channel: refactor force-close handling

aarani:


src/DotNetLightning.Core/Channel/ChannelHelpers.fs
```diff
@@ -17,3 +24,431 @@ module internal ChannelConstantHelpers =
...
+
+                let! perCommitmentSecret =
+                    match remotePerCommitmentSecret with
+                    | Choice1Of2 remotePerCommitmentSecret -> Ok remotePerCommitmentSecret
+                    | Choice2Of2 remotePerCommitmentSecretStore ->
+                        let commitmentSecretOpt =
+                            remotePerCommitmentSecretStore.GetPerCommitmentSecret commitmentNumber
+
+                        match commitmentSecretOpt with
+                        | Some commitmentSecret -> Ok commitmentSecret
```
aarani, line 261: We check for revoked closure after all our other checks return false, this makes it possible that a non-remote tx (maybe an old local commitment tx) gets caught by this function, so is it possible that this causes a False positive? (maybe there's a revoked remote commitment with the same commitment number)

src/DotNetLightning.Core/Channel/ChannelHelpers.fs
```diff
@@ -17,3 +24,431 @@ module internal ChannelConstantHelpers =
...
+
+                        transactionBuilder.Extensions.Add(CommitmentToLocalExtension())
+
+                        transactionBuilder
+                            .AddKeys(revocationPrivKey.RawKey())
+                            .AddCoin(ScriptCoin(closingTx, toLocalIndex |> uint32, toLocalScriptPubKey))
+                        |> ignore)
+
+                // FIXME: what if all the money is in htlcs?
+                if toLocalIndexOpt.IsNone && toRemoteIndexOpt.IsNone then
```
aarani, line 329: This is a temporary fix for the false-positive scenario, this causes the function to return "Unknown Closing Tx" if we couldn't find any to_local and to_remote output, but this creates an issue with commitment TXs that only have HTLC outputs and not any main one.

src/DotNetLightning.Core/Channel/ChannelHelpers.fs
```diff
@@ -17,3 +24,431 @@ module internal ChannelConstantHelpers =
...
+                    MainOutput = Error UnknownClosingTx
+                }
+
+    let HandleFundingTxSpent
+        (savedChannelState: SavedChannelState)
+        (remoteNextCommitInfoOpt: Option<RemoteNextCommitInfo>)
+        (channelPrivKeys: ChannelPrivKeys)
+        (closingTx: Transaction) : ClosingResult =
+        let closingTxId = closingTx.GetTxId()
+        if closingTxId = savedChannelState.LocalCommit.PublishableTxs.CommitTx.Value.GetTxId() then
```
aarani, line 430: Is there any other way to recognize local commitments? (This current check only works for last local commitment and not previous/revoked ones)

src/DotNetLightning.Core/Channel/ChannelHelpers.fs
joemphilips: I want you to add test to check the false positive scenario.

>  but this creates an issue with commitment TXs that only have HTLC outputs and not any main one.

And a test to check that this won't lead to the crash?

src/DotNetLightning.Core/Channel/ChannelHelpers.fs
joemphilips: 🤷

src/DotNetLightning.Core/Channel/ChannelHelpers.fs
```diff
@@ -17,3 +24,431 @@ module internal ChannelConstantHelpers =
...
+                | claimMainOutput ->
+                    {
+                        MainOutput = claimMainOutput
+                    }
+            | _ ->
+                {
+                    MainOutput = Error UnknownClosingTx
+                }
+
+    let HandleFundingTxSpent
```
joemphilips, line 424: I think this function needs test?

joemphilips:
some @aarani 's questions are hard hard for me to answer. And I have no time to catch up. (Channel state handling is not my interest anyway.)
Since at least there is no malicious code, I'd better just approve this change after tests I've mentioned above.

src/DotNetLightning.Core/Channel/ChannelHelpers.fs
aarani: > I want you to add test to check the false positive scenario.
I added a test for old local commitment txs 
> And a test to check that this won't lead to the crash?
I think I should explain more, the only problem with that edge case is that instead of BalanceBelowDustLimit error we return UnknownClosingTx which seems a bad error for the case (I added a comment explaining it the code also)

src/DotNetLightning.Core/Channel/ChannelHelpers.fs
aarani: I added a test for revoked/remote/local cases.

knocte:
How's this looking Joe?

@canndrew hey hello, by any chance you can have a quick look in case you spot something off very quickly?

tests/DotNetLightning.Core.Tests/TransactionTests.fs
```diff
@@ -326,4 +329,551 @@ let testList = testList "transaction tests" [
...
+            NodeMasterPrivKey extKey
+        let remoteChannelPrivKeys = remoteNodeMasterPrivKey.ChannelPrivKeys (rand.Next(1, 100))
+        let remoteChannelPubKeys = remoteChannelPrivKeys.ToChannelPubKeys()
+
+        let fundingAmount = 10_000_000L |> Money.Satoshis
+        let fundingScriptPubKey =
+            Scripts.funding
+                localChannelPubKeys.FundingPubKey
+                remoteChannelPubKeys.FundingPubKey
+        let fundingDestination = fundingScriptPubKey.WitHash :> IDestination
```
joemphilips, line 352: It is not super-important, but I want you to have a small scope for temporary variable like this.
i.e., instead of writing
```fsharp
let x = 1
let y = 2
let z = x + y
```

write

```fsharp
let z =
    let x = 1
    let y = 2
    x + y
```

In this way I can understand the lifetime of `x`, `y` more easily and it gets easy to review. (This is one of the biggest reason I use F#)

In this case, `fundingScriptCode` is what we want to have later. So

```fsharp

let fundingScriptCoin =
    let fundingDestination = // ..
    // ...
    ScriptCoin(fundingCoin, fundingScriptPubKey)
```

knocte: love that! I wonder if we could write an FSharpLint rule that could suggest this

tests/DotNetLightning.Core.Tests/TransactionTests.fs
```diff
@@ -326,4 +329,551 @@ let testList = testList "transaction tests" [
...
+                remoteChannelPrivKeys
+                localCommitmentTx
+
+        let outputsSumFromRemoteSpendingLocalCommitment =
+            (Result.deref remoteSpendingLocalCommitmentRes.MainOutput)
+                .SendAll(remoteDestPubKey.WitHash.ScriptPubKey)
+                .SendFees(Money.Zero)
+                .BuildTransaction(true)
+                .Outputs
+                |> Seq.sumBy (fun output -> output.Value)
```
joemphilips, line 579: Must be better if you pick the one you intended instead of summing up?

```fsharp
let dest = remoteDestPubKey.WItHash.ScriptPubKey
// ...
    Seq.pick (fun output -> if output.ScriptPubKey = dest then Some (output.Value) else None) // instead of Seq.filter
```

tests/DotNetLightning.Core.Tests/TransactionTests.fs
```diff
@@ -326,4 +329,551 @@ let testList = testList "transaction tests" [
...
+            
+        let expectedAmountFromToLocal =
+            let localAmount = lastLocalCommitmentSpec.ToLocal.ToMoney()
+            let fee = revokedLocalCommitmentTx.GetFee [| fundingScriptCoin :> ICoin |]
+            localAmount - fee
+        let expectedAmountFromToRemote =
+            lastLocalCommitmentSpec.ToRemote.ToMoney()
+
+
+        Expect.equal outputsSumFromRemoteSpendingRevokedRemoteCommitment (expectedAmountFromToRemote + expectedAmountFromToLocal) "wrong penalty tx"
```
joemphilips, line 623: You are running 3 attempt for `HandleFundingTxSpent`, and setup and assertions are mostly the same, it can be more succinct.

tests/DotNetLightning.Core.Tests/TransactionTests.fs
aarani: Done.

tests/DotNetLightning.Core.Tests/TransactionTests.fs
aarani: Done.

tests/DotNetLightning.Core.Tests/TransactionTests.fs
aarani: I grouped validation, commitment tx creation (as much as I could) into functions

tests/DotNetLightning.Core.Tests/TransactionTests.fs
```diff
@@ -326,4 +329,497 @@ let testList = testList "transaction tests" [
...
+            LocalCommit = remoteLocalCommit
+            RemoteCommit = remoteRemoteCommit 
+            LocalChanges = LocalChanges.Zero
+            RemoteChanges = RemoteChanges.Zero
+        }
+
+        let validateClosingResult (expectedAmount: Money) (errorMsg: string) (result: ClosingHelpers.ClosingResult) =
+            let dest =
+                let remoteDestPrivKey = new Key()
+                let remoteDestPubKey = remoteDestPrivKey.PubKey
```
joemphilips, line 553: nit: you don't have to create variable `remoteDestPubKey` and just use`remoteDestPrivKey.PubKey.WitHash.ScriptPubKey` directly in case like this.
The only case you should assign a new variable like this is to give a new information to the reader by its name.
e.g. 

```fsharp
// good example
let pubkeyToUseInMultisig = myPrivKey.PubKey // it is not self-evident from the right hand that this key is going to be used for multisig

// bad example
let pubkeyToUseInMultisig = myPrivKeyForMultisig.PubKey // no new information
```

src/DotNetLightning.Core/Channel/ChannelHelpers.fs
```diff
@@ -17,3 +24,439 @@ module internal ChannelConstantHelpers =
...
+                toRemoteIndexOpt
+                |> Option.iter
+                    (
+                        fun toRemoteIndex ->
+                            let localPaymentPrivKey =
+                                perCommitmentPoint.DerivePaymentPrivKey
+                                    localChannelPrivKeys.PaymentBasepointSecret
+
+                            transactionBuilder.SetVersion TxVersionNumberOfCommitmentTxs
+                            |> ignore<TransactionBuilder>
```
joemphilips, line 303: nit: why not just `ignore` ?

aarani: We have a policy wrt this, the reason (short answer) I think is to not accidentally ignore a function (e.g. int -> TransactionBuilder) instead of the outcome/result of the function. @knocte can give you a more complete answer.
Ah also there's a tweet about this: https://twitter.com/ChetHusk/status/1351156012590526474

tests/DotNetLightning.Core.Tests/TransactionTests.fs
aarani: Fixed

src/DotNetLightning.Core/Channel/ChannelHelpers.fs
joemphilips: Ah I see, I think I will follow this rule too from now on.

tests/DotNetLightning.Core.Tests/TransactionTests.fs
joemphilips: > P.S: all these codes are taken from the same TransactionTests file so I didn't really pay more attention to this sort of stuff

I cannot believe I wrote code like this by myself ...

aarani: Yes, it was probably missed in previous reviews 👍🏻 
https://github.com/joemphilips/DotNetLightning/blob/master/tests/DotNetLightning.Core.Tests/TransactionTests.fs#L159

joemphilips (approved):
LGTM: a80670236314233253ef568e12705ce27f435402

knocte:
> LGTM: https://github.com/joemphilips/DotNetLightning/commit/a80670236314233253ef568e12705ce27f435402

Rebasing from github UI (no changes).

PR #185, opened 2022-04-15
Core: remove dead code

knocte:
These files were not included in the build.
(In fact 2 of them are empty and the 3rd is
incomplete, it doesn't even build).


knocte:
@joemphilips ping?

PR #186, opened 2022-04-24
add DotNetLightning.ClnRpc

joemphilips:
This commit adds a new project for calling c-lightning from .NET
in a typesafe way.
Most of the codes are auto-generated from client generator built in
ref: https://github.com/ElementsProject/lightning/pull/5201

It must be updated after https://github.com/ElementsProject/lightning/pull/5216 gets merged.
So that all the generation logics is defined in this repo.
But this is good enough for temporary solution.


joemphilips:
Before this get merged, I must review and merge https://github.com/joemphilips/DotNetLightning/pull/184
and do some cleanup...

knocte:
> I must review and merge https://github.com/joemphilips/DotNetLightning/pull/184

And after #184 I'll merge the other approved ones before this one, if you don't mind.

src/DotNetLightning.Core/DotNetLightning.Core.fsproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <VersionPrefix>1.1.2</VersionPrefix>
-    <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netstandard2.1;net6</TargetFrameworks>
```
knocte, line 4: I think the correct name for this is net6.0, not net6

src/DotNetLightning.Core/Utils/LNMoney.fs
```diff
@@ -50,10 +50,7 @@ type LNMoney = | LNMoney of int64 with
         LNMoney.MilliSatoshis(Checked.op_Multiply 1000L sats)
 
     static member inline Satoshis(sats) =
-        LNMoney.Satoshis(int64 sats)
-
-    static member Satoshis(sats: uint64) =
-        LNMoney.MilliSatoshis(Checked.op_Multiply 1000UL sats)
```
knocte, line 56: is this change related to this PR? if I had to choose one overload to leave here, I would leave the one that has the type explicitly specified

joemphilips: > is this change related to this PR?

No, I should create another PR.

>  if I had to choose one overload to leave here, I would leave the one that has the type explicitly specified

May I ask why?
I just deleted this because I'm having compile error, I thought it is good idea to have inline method instead of having overload for all integral types... But I found inline static methods are tricky and sometimes it complains that it cannot infer types.

I don't know which is better, to leave only inline method or the one explicitly specified.

src/DotNetLightning.Core/DotNetLightning.Core.fsproj
knocte: actually maybe I'm wrong

src/DotNetLightning.Core/Utils/LNMoney.fs
knocte: well let's move it to a new PR indeed, and we discuss there?

knocte:
Not sure why the mono lane failed, maybe just re-run it?

knocte:
To prevent the `style` CI lane from failing on auto-generated files, you can add those files to a `.fantomasignore` file.

Issue #58, opened 2020-02-28
TODO: Code Formatting


joemphilips:
Right now we don't have any automated code formatting scheme.
It is nice to have formatting pipeline in CI.

knocte:
Via fantomas?

joemphilips:
@knocte probably, I don't know what is the state of the art.

knocte:
I think fantomas is the state of the art, if not the only option. But I wouldn't make use of it at least until this PR is merged: https://github.com/fsprojects/fantomas/pull/649 otherwise the inconsistency between having a space before parenthesis or not becomes too annoying.

joemphilips:
Thanks for the info. I think it still has some benefits without that feature, but there is no hurry.

knocte:
I don't know, to me, having different spacing style depending on if the function called/defined starts with an uppercase letter or not, is master-level of bikeshedding.

PR #188, opened 2022-04-26
Apply automatic source-code formatting for F#

knocte:
Fixes https://github.com/joemphilips/DotNetLightning/issues/58


.github/workflows/ci.yml
```diff
@@ -48,3 +48,21 @@ jobs:
...
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - name: Setup .NET Core SDK 5.0.x
+        uses: actions/setup-dotnet@v1.7.2
+        with:
+          dotnet-version: '5.0.x'
+      - name: fantomless
```
joemphilips, line 62: Looks great, but why not use fantomas and use `fantomless`?

joemphilips: Oh nevermind I found https://github.com/nblockchain/fantomless#why-a-fork

joemphilips:
I think we should use fantomas instead of fantomless.
IMO, the standardization matters more than the beauty/ergonomics of the code when it comes to code formatting.
You can bump the fantoms version when those changes proposed (e.g. https://github.com/fsprojects/fantomas/issues/2111)
 gets resolved.

knocte:
> IMO, the standardization matters more than the beauty/ergonomics

In general, I agree with this statement.

But in particular, in this case, the main reason we made fantomless is that if you read the conversations on those 3 github issues that fantomless implements, you'll see the fantomas maintainer advocating for adopting an F# behaviour that is only compatible with F#6.0 or higher, which would mean basically rendering DNL incompatible with older versions of F#. That's why we didn't propose a PR for that particular issue (and hence it's stalled).

.github/workflows/ci.yml
```diff
@@ -48,3 +48,21 @@ jobs:
         dotnet restore -p:Portability=True DotNetLightning.sln
         msbuild src/DotNetLightning.Core/DotNetLightning.Core.fsproj -p:Portability=True -p:TargetFramework=netstandard2.0
 
+  style:
+    needs: build_and_test
```
joemphilips, line 52: I think this should run before testing.
Since it is more likely to bump in, and it is easier to fix than unit tests.

knocte: Mmm, are you sure? I actually thought about the order before, and I've always decided to put style the last, because we don't want to annoy the contributor about style fixes until he has got the other (most) important things right.

joemphilips:
Well, alright. Having some kind of style is better than no style at all, and I don't want to get too deep into the vanity indentation issue. So I guess This PR is OK after the test order I've mentioned above gets fixed.

BTW, do you remember you assured me that you will stop using custom `Result` type in ResultUtils when debian packaging supports F# 4.5+ ? https://github.com/joemphilips/DotNetLightning/pull/143#discussion_r570731406
Isn't ready yet?

knocte:
> Isn't ready yet?

Not ready yet, but I think I'm going on a different route. If in 2 months I haven't removed the ResultType hack, ping me again.

.github/workflows/ci.yml
joemphilips: hmm, I see. Well, then run those 2 in parallel? (that is, drop `needs`)

knocte: > (that is, drop needs)

Ok sure I can do that; but this doesn't remove the "annoying the contributor too early" part :)

joemphilips: contributor gets annoyed by CI anyway. Thats what CI is for.

knocte: Done.

joemphilips:
lgtm 115d7fd6f5f79de3ef58163bcec0223593c984d8

joemphilips (approved):

knocte:
@joemphilips how come the publish CI lane failed while the main one did not?

knocte:
Ok I found out why:
```
  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
    <WarningLevel>3</WarningLevel>
    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
  </PropertyGroup>
```
This means that our non-publish CI lanes are only building in DEBUG mode, we should fix that.

And tomorrow I'll look at those new warnings.

joemphilips:
ohh... Thanks, I will look in too.

PR #189, opened 2022-04-26
wip: test publish works fine in .NET SDK 6.x

joemphilips:


knocte:
> run ci in Release build

And I'm not sure about that either. I think tests should be run in DEBUG because if a problem happens, you want verbose stack traces and the nicest things that DEBUG can bring for you to solve the bug.

joemphilips:
> And I'm not sure about that either. I think tests should be run in DEBUG because if a problem happens, you want verbose stack traces and the nicest things that DEBUG can bring for you to solve the bug.

You are correct again... I guess I need some nap.

PR #191, opened 2022-04-28
some package cleanups.

joemphilips:
* [x] stop using obsolete netcoreapp3.0
* [x] fix: https://github.com/joemphilips/DotNetLightning/issues/59
* [ ] ~~Update some doc comments and use `fsdoc` for document generation?~~
* [ ] ~~Stop using prefix versioning and go beta?~~ -> moved to https://github.com/joemphilips/DotNetLightning/pull/192

PR #194, opened 2022-05-01
GitHubCI: merge both publish_* workflows into one

knocte:
A recent PR tweaking the publish workflows had to duplicate a job
because the 'needs' element doesn't seem to work across workflows.
By merging these two workflows into one, but separating them via jobs,
we can fix this problem so that we get less duplicated code. And the
CI code for publishing to Nuget is very similar anyway, so we might
reuse more lines in the near future as well, there was no need to
separate them.

NOTE: Now the 'on' element of the workflow contains both branches and
tags, but each job checks for what it needs via an 'if' check that
contains a "startsWith" clause, explained here:
https://github.community/t/if-not-startswith-mutually-exclusive-steps/141841/2

And the way to combine two 'if' elements is explained here:
https://github.community/t/how-to-write-multi-line-condition-in-if/128477/8


joemphilips:
LGTM
nit: It is irrelevant to this PR. but `pack_and_push_prerelease` would better bump the .NET SDK Version to 6.0.

knocte:
> but pack_and_push_prerelease would better bump the .NET SDK Version to 6.0.

Shouldn't that be done in the PR where you have this change? BTW, I just looked at the relevant commit that does this in that PR and you were indeed forgetting to change it in publish_release.yml. So I guess it's better you rebase that PR first.

joemphilips:
> Shouldn't that be done in the PR where you have this change? BTW, I just looked at the relevant commit that does this in that PR and you were indeed forgetting to change it in publish_release.yml. So I guess it's better you rebase that PR first.

Yes, I will do it in https://github.com/joemphilips/DotNetLightning/pull/191

PR #195, opened 2022-05-05
Fix compile warning on non-portability build.

joemphilips:
The compiler were warning in non-Portability build
that it has a version mismatch in `System.Runtime.CompilerServices.Unsafe`

This happens only in netstandard2.0 and not in 2.1.

This seems to be the similar issue to https://github.com/dotnet/runtime/issues/46082
that `System.Memory` in netstandard2.0 has a transitive dependency to
`System.Runtime.CompilerServices.Unsafe`.

So fix it by adding Unsafe explicitly in projects using it, as suggested
in the issue above.

### DIFF
--- a/src/Macaroons/Macaroons.csproj
+++ b/src/Macaroons/Macaroons.csproj
@@ -6,7 +6,7 @@
     </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\NSec\Experimental\NSec.Experimental.csproj" PrivateAssets="all" />
+    <ProjectReference Condition="'$(BouncyCastle)'!='true'" Include="..\NSec\Experimental\NSec.Experimental.csproj" PrivateAssets="all" />
   </ItemGroup>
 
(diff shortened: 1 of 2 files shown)


knocte:
If this was only an issue on the nonPortability build, how about adding a `Condition="'$(BouncyCastle)'!='true'"` attrib to the PackageReference?

joemphilips:
> If this was only an issue on the nonPortability build, how about adding a Condition="'$(BouncyCastle)'!='true'" attrib to the PackageReference?

I have updated it to fix the issue in more cleaner way.
(The root cause of the problem was that `NSec/Cryptography` was referencing the `Unsafe` in its csproj which was unnecessary)

knocte:
LGTM, except the 3rd(last) commit which is unrelated to this PR, and is not a nit (it introduces new C# syntax).

joemphilips:
Merged after dropping the 3rd commit.

PR #197, opened 2022-05-12
FsharpLint support

janus:


joemphilips (approved):
Thanks @janus this looks nice!
Can I merge now or do you want to add something?

PR #198, opened 2022-05-12
Core: fix bug in local force close handling

aarani:
Local commitment txs use RemoteParams.ToSelfDelay,
using LocalParams.ToSelfDelay when searching for outputs
can cause problem if:
LocalParams.ToSelfDelay <> RemoteParams.ToSelfDelay

### DIFF
--- a/src/DotNetLightning.Core/Channel/ChannelHelpers.fs
+++ b/src/DotNetLightning.Core/Channel/ChannelHelpers.fs
@@ -202,7 +202,7 @@ module ClosingHelpers =
                 let toLocalScriptPubKey =
                     Scripts.toLocalDelayed
                         remoteCommitmentPubKeys.RevocationPubKey
-                        staticChannelConfig.LocalParams.ToSelfDelay
+                        staticChannelConfig.RemoteParams.ToSelfDelay
                         localCommitmentPubKeys.DelayedPaymentPubKey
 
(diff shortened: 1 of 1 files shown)


joemphilips (approved):

knocte:
@aarani hey, sorry I forgot this nit in this PR: you say "fix bug" and "can cause problem", but you don't describe the problem. Can you describe it here for the record please?

aarani:
> @aarani hey, sorry I forgot this nit in this PR: you say "fix bug" and "can cause problem", but you don't describe the problem. Can you describe it here for the record please?

Using `LocalParams.ToSelfDelay` for generating the scriptPubKey, makes a different scriptPubKey than the one used in the commitment tx which causes the code to not be able to find the to_local output. (The code ignores the commitment tx as BalanceBelowDustLimit)

knocte:
> causes the code to not be able to find the to_local output. (The code ignores the commitment tx as BalanceBelowDustLimit)

And what would be the consequence of that from the user's PoV?

aarani:
> > causes the code to not be able to find the to_local output. (The code ignores the commitment tx as BalanceBelowDustLimit)
> 
> And what would be the consequence of that from the user's PoV?

Not being able to create `recovery tx` for their own output (if they initiate the force close).

PR #200, opened 2022-05-19
Fix doc build

joemphilips:


docs/c-lightning/overview.md
```diff
@@ -1,8 +1,8 @@
 ---
-title: calling c-lightning rpc and building plugins.
-category: c-lightning support
+title: overview
+category: c-lightning
```
knocte, line 3: @joemphilips I think CLightning project got renamed to LightningCore

docs/c-lightning/overview.md
```diff
@@ -1,8 +1,8 @@
 ---
-title: calling c-lightning rpc and building plugins.
-category: c-lightning support
+title: overview
```
knocte, line 2: nit: Overview instead of overview, to be in sync with other titles?

docs/c-lightning/overview.md
joemphilips: Actually its "core lightning" I believe this is to keep compatibility with old "c-lightning"
but yeah why not. I will rename it to "core-lightning".

joemphilips: Actually I use the term "c-lightning" all over the place already.
I think I will keep it as is.

PR #212, opened 2022-06-23
Core: add LTC network prefixes

aarani:
This commit adds support for creating LTC invocies.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -29,7 +29,7 @@
   </Choose>
 
   <ItemGroup>
-    <PackageReference Include="NBitcoin" Version="6.0.19" />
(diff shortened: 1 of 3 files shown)


joemphilips:
You need test

aarani:
> You need test

Added.

knocte:
LGTM

joemphilips:
Code change itself seems good.

I'm wondering if we should remove the dependency to `NBitcoin` and solely rely to `NBitcoin.Altcoins`.
having both as a deps might cause a version mismatch issue when bumping etc...
Do you have any opinion on this?

knocte:
> Do you have any opinion on this?

We tested your suggested approach and it works, so the PR has been updated. Merge? Review? :) Thanks

joemphilips:
OK I will just merge it for now.

PR #214, opened 2022-07-12
Core/Primitives: less circular type deps to workaround fantomas bug

knocte:
In recent commit to adjust formatting [1], an #if clause was removed,
resulting in applying the `[<Struct>]` attrib always. This is a fantomas
bug which I just reported here [2], but if we reduce circular type
dependencies here, we can just workaround it for now.

[1] https://github.com/joemphilips/DotNetLightning/commit/c0747c951a5b9cfd94180102346ae3419b00f57a
[2] https://github.com/fsprojects/fantomas/issues/2361

### DIFF
--- a/src/DotNetLightning.Core/Utils/Primitives.fs
+++ b/src/DotNetLightning.Core/Utils/Primitives.fs
@@ -43,6 +43,63 @@ module Primitives =
 
             output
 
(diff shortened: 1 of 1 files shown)


joemphilips (approved):
LGTM

PR #221, opened 2022-08-04
fix bug in ClnClient.SendCommandAsync

joemphilips:


knocte:
what bug?

joemphilips:
@knocte I forgot to use custom JsonSerializer when serializing the request.
Pretty obvious.

knocte:
> @knocte I forgot to use custom JsonSerializer when serializing the request.

That's not the bug, that's the bugfix :) And you didn't include neither in the commit msg.

joemphilips:
> And you didn't include neither in the commit msg

Ok, I've updated the commit msg.

knocte:
> Ok, I've updated the commit msg.

For the record, the commit msg still explained the bugfix but not the bug :)
What should have been added is what was the consequence of JSON not following default settings? That would be the bug ;)

PR #222, opened 2022-08-05
simplify PluginServerBase thread safety

joemphilips:
Cln Plugin must never write to stdout concurrently, otherwise its
message will be corrupted.
Before this commit, `PluginServerBase` were using
`AsyncSemaphore` to ensure the thread safety for stdout.
This is hard to understand, especially when an user side and a
DNL side both tries to get lock (e.g. for `SendNotification`) and
cause deadlock.

This commit simplifies it by just using `Stream.Synchronized`
to assure the thread safety.

### DIFF
--- a/examples/HelloWorldPlugin/HelloWorldPlugin.cs
+++ b/examples/HelloWorldPlugin/HelloWorldPlugin.cs
@@ -1,9 +1,4 @@
-using System;
-using System.Collections.Generic;
-using System.Threading.Tasks;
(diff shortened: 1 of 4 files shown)


joemphilips:
I've accidentally committed this change into master 92588224bc4bad8d913a086780ff36b75c3717fb 😓
reverted in 9d6a2c68dd272fc9706c633930b8f268418b627c

Issue #224, opened 2022-08-14
Nuget package for DotNetLightning.ClnRpc has missing dependencies


rsafier:
Missing two packages which are not public on Nuget so cannot try sample plugin code:
![image](https://user-images.githubusercontent.com/645964/184528221-89b32b35-049a-43e5-bce6-9a6e2fa707bf.png)

knocte:
Hello @rsafier , thanks for reporting this problem!

I believe we have hit this problem before with the DotNetLightning nuget package, and we fixed it this way: https://github.com/joemphilips/DotNetLightning/commit/c98307465f647257df1438beadb4cabc7db757f2 .

So I guess we now need the same workaround for the DotNetLightning.ClnRpc project. Can you create a PR for this?

joemphilips:
@rsafier 
You are referencing old version of `DotNetLightning.ClnRpc` The newest version is defined by prefix version number. (in the same way with [`System.Commandline` beta version](https://github.com/dotnet/command-line-api/tags) management.)

This is the newest.
https://www.nuget.org/packages/DotNetLightning.ClnRpc/1.1.2-date20220806-0225-git-7c5fdbb

@knocte I believe it is a different problem. the latest version has no dependency for ResultUtils and PluginLogger

Recently I received a similar feedback from others, it seems that managing a package version with VersionPrefix is not user-friendly.
I must consider defining a publication flow with semantic versioning... probably with the way similar to NBitcoin.

knocte:
I guess it wasn't really @rsafier's fault because the version of the nuget package that fixed this issue was a pre-release. But I see that you just published a stable version (1.1.6) that contains the fix, so I guess we can close this now.

joemphilips:
Yeah, I guess I will create a new release manually every time there is a change so that this kind of confusion won't happen again.

At some point I should introduce more sophisticated/automated flow.

PR #225, opened 2022-08-15
cln: add JsonConverter for handling enum

joemphilips:
Unlike Newtonsoft.Json, System.Text.Json does not
respect EnumMemberAttribute. (see: https://github.com/dotnet/runtime/issues/31081)
This caused an issue when sending request to c-lightning,
when you use System.Text.Json for serialization lib, and there
is a enum value in the request field, the field will be serialized into
an int, instead of a string. Thus causing RPC error.

This commit fixes it by using `JsonStringEnumConverterEx`
(which is taken from an issue comment of issue above).

### DIFF
--- a/.github/workflows/build+test.yml
+++ b/.github/workflows/build+test.yml
@@ -62,7 +62,7 @@ jobs:
       - name: Setup .NET 6
         uses: actions/setup-dotnet@v1
         with:
-          dotnet-version: 6.0.x
+          dotnet-version: 6.0.303
 
       - name: build project assemblies so that fsdocs can read assemblies and .xml files from bin/ .
         run:
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Setup .NET 6
         uses: actions/setup-dotnet@v1
         with:
-          dotnet-version: 6.0.x
+          dotnet-version: 6.0.303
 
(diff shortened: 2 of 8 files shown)


knocte:
I see you're trying to find the packages dir where nuget packages get downloaded. I recently had trouble with this because apparently M$ has changed behaviour; to go back to the old behaviour of creating a packages subfolder in the repo itself when restoring, you can create a NuGet.config file with this content:

```
<configuration>
    <config>
        <add key="globalPackagesFolder" value=".\packages\" />
    </config>
</configuration>
```

joemphilips:
@knocte Thanks, but it seems this is a different issue.

It seems that fsdocs fails to find a path to 3rd party dll, (in this case NBitcoin)
ref: https://github.com/fsprojects/FSharp.Formatting/issues/616

I don't know why but it works in local, event when I remove the nuget cache with `dotnet nuget locals global-packages -c`

I should probably change the publish workflow to more manual way, as I have to do it anyway for https://github.com/joemphilips/DotNetLightning/issues/224

joemphilips:
> It seems that fsdocs fails to find a path to 3rd party dll, (in this case NBitcoin)
ref: https://github.com/fsprojects/FSharp.Formatting/issues/616
> I don't know why but it works in local, even when I remove the nuget cache with dotnet nuget locals global-packages -c

@knocte  hmm, it works fine when I use an old version of .NET SDK. Maybe the root cause is same with yours.
Anyway I should report this to fsdocs repo.

PR #228, opened 2022-08-23
Serialization: fix SetFeature bug

knocte:
This removes the use of intermediate bytearray when cloning the feature bitarray, it also adds a test that can prove this problem.


joemphilips (approved):
LGTM

joemphilips:
pushed new version v1.1.9

PR #231, opened 2022-08-25
Add DnsHostname type to NetAddress

joemphilips:
DNS hostname type has been added to node_announcement msg.
This commit supports is serialization.


joemphilips:
This PR also changes the NetAddress serialziation logic, since according to the spec we should not be as strict as it was.
Thanks for reporting @aarani !

joemphilips:
Pushed as v1.2.0

PR #232, opened 2022-08-30
Core(Serialization): made UnsignedNodeAnnouncementMsg.Features a Result

webwarrior-ws:
Changed type of `Features` field in `UnsignedNodeAnnouncementMsg` to
`Result<FeatureBits, FeatureError>` in order to handle cases with invalid
or unknown features without throwing an exception during deserialization.

### DIFF
--- a/src/DotNetLightning.Core/Serialization/Msgs/Msgs.fs
+++ b/src/DotNetLightning.Core/Serialization/Msgs/Msgs.fs
@@ -1357,12 +1357,12 @@ and UnknownNetAddr = byte
 
 /// `node_announcement` in [bolt07](https://github.com/lightning/bolts/blob/master/07-routing-gossip.md)
 /// without signatures.
-/// Features is Result<FeatureBits, FeatureError> because node_announcement with invalid or unknown features
-/// should not raise exception on deserialization.
+/// Features are stored as byte array because node_announcement with invalid or unknown features
+/// should not raise exception on deserialization. They can be accessed using Features property
 [<CLIMutable>]
 type UnsignedNodeAnnouncementMsg =
     {
-        mutable Features: Result<FeatureBits, FeatureError>
+        mutable FeatureBitsArray: array<byte>
         mutable Timestamp: uint32
         mutable NodeId: NodeId
         mutable RGB: RGB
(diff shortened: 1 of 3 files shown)


joemphilips:
I don't really like having `Result` inside of DTOs.

Why don't you
1. change the field to `FeatureBitArray: BitArray` (or `FeatureBitsArray: byte[]`)
2. and add `member this.FeatureBit = this.FeatureBitArray |> FeatureBits.TryCreate`

That is more friendly for who wants to serialize DTOs as-is.

Also, I wan't you to add test. It seems that you can not serialize the FeatureBits in the same way it has come in in the current code.

src/DotNetLightning.Core/Serialization/Msgs/Msgs.fs
```diff
@@ -1357,12 +1357,12 @@ and UnknownNetAddr = byte
...
 /// without signatures.
-/// Features is Result<FeatureBits, FeatureError> because node_announcement with invalid or unknown features
-/// should not raise exception on deserialization.
+/// Features are stored as byte array because node_announcement with invalid or unknown features
+/// should not raise exception on deserialization. They can be accessed using Features property
 [<CLIMutable>]
 type UnsignedNodeAnnouncementMsg =
     {
-        mutable Features: Result<FeatureBits, FeatureError>
+        mutable FeatureBitsArray: byte []
```
knocte, line 1365: @webwarrior-ws we prefer the notation `array<byte>`

knocte:
> Core(Serialization): changed how features are stored in UnsignedNodeA…
> …nnouncementMsg

Nit: commit msg title is too long, so it gets split.

tests/DotNetLightning.Core.Tests/Serialization.fs
```diff
@@ -108,8 +108,12 @@ module SerializationTest =
                                                 "03f3c15dbc4d425a4f4c36162a9159bb83511fa920dba1cc2785c434ecaf094015"
                                             )
                                         )
-                                    Features =
-                                        FeatureBits.CreateUnsafe [| 0uy |]
+                                    FeatureBitsArray =
```
joemphilips, line 111: The test case must handle the feature bit with unsupported mandatory flag.

tests/DNLTestBase/Generators/Msgs.fs
```diff
@@ -588,7 +590,7 @@ let unsignedNodeAnnouncementGen =
 
         return
             {
-                Features = f
+                FeatureBitsArray = f.ByteArray
```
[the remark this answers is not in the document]
webwarrior-ws: done

PR #236, opened 2022-10-13
No mutable fields in message records

### LINKED ISSUE #87: Deserialize method should not return `unit`
The consequence of this is that the types need their members to be `mutable`, which is ugly.

webwarrior-ws:
Fixes #87 
Made `ILightningSerializable<'T>.Deserialize` method return
 deserialized object instead of mutating `this` instance.
Removed `mutable` modifier from all record fields in message types.


joemphilips (approved):
Thanks. LGTM. Interesting hack.
You are correct that all these mess in deserialization is caused by the lack of static abstract members. But probably we cannot use it in the future even after F#7 gets stable, for the backward compatibility reason.

PR #238, opened 2023-04-18
CI: fix fsDocs

aarani:
This commit pins the version for fsdocs-tool to
v16.1.1.


knocte:
Missing the why ("fix fsDocs" is not really explaining the why unless you explain what you're fixing).

